Issue #1, opened 2017-05-19
supports snapshot methods


sculove:
## Description
egjs-infiniteGrid supports snapshot method like `setStatus` and `getStatus` for using persist
I would like to provide a snapshot method at Flicking component like infiniteGrid.

ref : https://github.com/naver/egjs/issues/443

PR #5, opened 2017-06-21
feat(flicking): Intent to ship new status methods

netil:
Implement method to get/set snapshot of current flicking contents

Ref #1

### DIFF
--- a/src/flicking.js
+++ b/src/flicking.js
@@ -634,18 +634,21 @@ export default class Flicking extends Mixin(Component).with(EventHandler) {
 
 		const $first = $panel.querySelector(":first-child");
 
-		height = $first.getAttribute(consts.DATA_HEIGHT);
+		if ($first) {
+			height = $first.getAttribute(consts.DATA_HEIGHT);
 
-		if (!height) {
-			$children = $panel.children;
-			height = utils.outerHeight(
-				$children.length > 1 ? ($panel.style.height = "auto", $panel) : $first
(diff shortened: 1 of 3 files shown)


src/flicking.js
```diff
@@ -1318,6 +1321,68 @@ export default class Flicking extends Mixin(Component).with(EventHandler) {
...
+			panel: {
+				index: panel.index,
+				no: panel.no,
+				currIndex: panel.currIndex,
+				currNo: panel.currNo
+			},
+
+			// panel's html
+			$list: panel.$list.map(v => ({
+				style: v.style.cssText.replace(rxStyle, "").trim(),
```
sculove: - Do you have to check this out? `cssText` of panel isn't restore.
```js
v.style.cssText.replace(rxStyle, "").trim()
```

- Does `cssText` of base element is not save?

netil: Thinking the whole process of getting and restoring will be like:
- on getting: instantiate -> some user action(or not) -> get current status of flicking element
- on restoring: instantiate -> restore the previous state with state data

Will not save whole css style, because on restoring the flicking element can stay differently than the previous (ex. get status when is vertical mode, and user changes the portrait mode of device).

To avoid that situation, keeping the instantiated style will be more reasonable, rather than overwriting with the previous.

sculove: Ok I got it.
LGTM

Issue #23, opened 2017-08-14
unexpected restore event firing


netil:
## Description
`restore` custom event is fired inappropriately in specific occasion.

## Steps to check or reproduce
- Do touch moves of panel, but to be restored
- Call `.next/prev/moveTo` with duration
- unexpected `restore` event is fired after `flickEnd` event

PR #26, opened 2017-08-14
fix(Flicking): Correct wrong restore event calls

netil:
## Issue

#23 

## Details

Fix to not be called restore event in normal flow, after restoration happening


sculove:
LGTM

PR #45, opened 2017-09-26
feat(event): Intent to ship event.isTrusted

netil:
## Issue

#41

## Details

Implement flag value to determine if the event was started by explicit user's action or not.


src/eventHandler.js
```diff
@@ -10,10 +10,12 @@ export default superclass => class extends superclass {
 	 */
 	_holdHandler(e) {
 		const conf = this._conf;
+		const touch = conf.touch;
 		const holdPos = e.pos.flick;
 
-		conf.touch.holdPos = holdPos;
-		conf.touch.holding = true;
+		touch.holdPos = holdPos;
+		touch.holding = true;
```
sculove: eg.Axes v2.2.0 pass holding and isTrusted properties.
how about using these properties?

netil: May be it's good, but using prop value from the dependency will make deep attachment and is likely have unknown error if the dependency module changes its behavior.

netil:
isTrusted is added for all events. 
- https://github.com/netil/egjs-flicking/blob/2ade44a605da8364ece43cf5b56f8012a4642cdd/src/Flicking.js#L883

sculove:
LGTM

Issue #66, opened 2018-01-05
Remove outdated code


netil:
## Description

Due to increase of performance device usages, function to determine `isHWAccelerable` isn't useful anymore.
Remove from the code base.

Ref #42

PR #67, opened 2018-01-05
style(utils): Remove outdated .isHWAccelerable()

netil:
## Issue

#66

## Details

Remove unused & outdated code

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ before_install:
   - yarn global add greenkeeper-lockfile@1
 install:
   - npm install
+sudo: required
 addons:
   chrome: stable # install chrome stable
 cache:
--- a/src/config.js
+++ b/src/config.js
@@ -39,7 +39,7 @@ const CONFIG = {
 
 // default options
 const OPTIONS = {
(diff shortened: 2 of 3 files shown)


sculove:
@netil the default value of hwAccelerable is true.
How about changing default value of hwAccelerable option?
eg.Flicking should support android 2.3+
so, How about changing default value of hwAccelerable option from 'true' to the result value of the isHWAccelerable function?

netil:
@sculove I don't think so.
The usage environment was upgraded comparing years ago. I know that there're still low efficiency devices out there, but we need to cut the support in certain time.

sculove:
@netil ok. you'd better correct browser coverages

sculove:
LGTM

netil:
@sculove I think it's good to have some discussion about it.

netil:
Adding code snippet url for the removed code will be enough

netil:
To check the support of CSS transform and HW acceleration, check out the below snippet
- https://gist.github.com/netil/c1d1b67e1c1926364c0709f110a09b44

Issue #64, opened 2018-01-03
diatance value


jadujadu:
onflick 이벤트에서
moveTo, next, prev 메소드 호출을 통해 전환시 distance가 pos와 동일한 값이 떨어집니다.
이후 수동으로 flicking 후 호출시 정상적인 값이 떨어집니다.

netil:
Hi @jadujadu 

The distance value is calculated as:
- distance = destination position - user touch position 

If there's no touch position value, then touch position value is treated as 0.
This is the reason why distance and position are same.

What is your criteria of the distance value when user moved panel without touch?

netil:
Internally we discussed the issue, the `distance` property is given from the touch start point value.
If it's called by API methods, there's no starting point value.
I think in this case, the distance value can't be calculated.

So, the `distance` prop value called by API should be `null`.

PR #68, opened 2018-01-11
fix(event): Set to null e.distance on API call

netil:
## Issue

#64

## Details

- Without an implicit user's action, the event's distance value should be set to null
- Update some dependency modules to the latest


sculove:
LGTM

Issue #61, opened 2017-12-28
Remove physical panel number terminology


netil:
## Description

In some circumstances, panels are duplicated due to maintain the usability.
Ex. when there's a panel, to make flickable, it duplicates to be 3.
In that case, logically there's a panel, but physically there're 3 panels.

This approach was quite reasonable at the first time, but it's giving confusion and isn't really used normally.
Get rid the confusion `physical` on API calls.

PR #71, opened 2018-01-12
style(Flicking): Mark physical as deprecated

netil:
## Issue

#61

## Details

Mark param as deprecated, which will be removed in the next 4th release.


sculove:
LGTM

PR #74, opened 2018-01-16
chore(flicking.d.ts): Update to extends Component

UYEONG:
## Details

There is a problem i TS compilation because `Flicking` did not inherit the parent class `Component`.


netil:
@UYEONG thanks for the contribution!

Issue #19, opened 2017-08-03
Intent to ship 'previewPadding' in percent value.


sculove:
@netil commented on [Thu Aug 11 2016](https://github.com/naver/egjs/issues/359)

## Description

For `previewPadding` options, only `px` values are allowed.
To be responsible, it should be act on `%`(relative wrapper element) values too.
## Steps to check or reproduce

``` js
 new eg.Flicking("#mflick", {
    previewPadding: [ "10%", "20%" ],
    circular: true
});
```

---

@netil commented on [Mon Nov 14 2016](https://github.com/naver/egjs/issues/359#issuecomment-260293585)

There's ambiguity on this.
If percentage values are given, what that values are based on?
- Wrapper element
- screen
- viewport (in case of desktop this will be browser's size)

---

@netil commented on [Thu Nov 17 2016](https://github.com/naver/egjs/issues/359#issuecomment-261189678)

As a result on internal discussion,
- `wrapper` is the base element 
- isn't must functionality for this moment, but will be useful providing

Keep for later shipment

---

@netil commented on [Fri Mar 03 2017](https://github.com/naver/egjs/issues/359#issuecomment-283850453)

Try implement on v2

PR #80, opened 2018-02-05
feat(options): Intent to ship previewPadding with '%' value

netil:
## Issue

#19

## Details

Make to accept '%' unit value in addition.


netil:
It can accept any units accepted for the padding value.
Describing it on API doc.

sculove:
LGTM. Unfortunately, coverage has fallen. ^^;

Issue #77, opened 2018-01-21
A bug occurs when executing`resize` method


JaeYeopHan:
## Description
A bug occurs when executing`resize` method in this module.

- Related Link: https://m.search.naver.com/search.naver?where=m&sm=mtb_etc&mra=blA2&query=%EC%97%AC%EB%A6%84%EA%BD%83

### Details
![eg-flick-conatiner_capture_image](https://user-images.githubusercontent.com/17924127/35191439-d804dea8-febd-11e7-945b-6b6be3cc2b55.png)
![eleemt-of-panel_capture_image](https://user-images.githubusercontent.com/17924127/35191440-da4f2ba0-febd-11e7-9dfe-bbfaa6cb987e.png)

When resize occurs, reassigning the width value of the container. If the panels already constituting the container are responsive, the reassignment occurs, and the layout of the elements constituting the panel is not properly detected.

## Steps to check or reproduce
1. Click the [link](https://m.search.naver.com/search.naver?where=m&sm=mtb_etc&mra=blA2&query=%EC%97%AC%EB%A6%84%EA%BD%83)
2. Open the **Chrome Development Tool**
3. Click the `rotate` button.

#### Current behavior
![current_behavior](https://user-images.githubusercontent.com/17924127/35191469-aa55da2e-febe-11e7-949c-52c31f17b528.png)

#### Expected behavior
![expected_behavior](https://user-images.githubusercontent.com/17924127/35191472-ae76f55c-febe-11e7-83a8-9840c2fb79db.png)

4. Click the rotate button again.
#### Current behavior
![current_behavior_2](https://user-images.githubusercontent.com/17924127/35191485-fe69506e-febe-11e7-8b59-8c087bbe2330.png)

#### Expected behavior
![expected_behavior_2](https://user-images.githubusercontent.com/17924127/35191486-00f3ba5e-febf-11e7-9080-f8cdb9a8c2d0.png)

What about adding an option value for the responsive case?
Thx.

netil:
Thanks @JaeYeopHan for the report. I'll try to check the issue.

netil:
It happens because of the padding set on container element.
When `.resize()` is called, it updates container element's width as the total width value of panel's width.

In that case, having a percentage value, it will having based on the container's width value, which is causing bigger padding rather than the viewport's width.

In the next release, will remove updating container's width.

For now to fix your case, it needs update padding percentage value as follow.
```css
[...]

PR #82, opened 2018-02-13
fix(resize): Remove updating container's width

netil:
## Issue

#77

## Details

There's no reason update container's width value as total panel's width.
It keeps to be 100% width to prevent styling issue.


netil:
The coverage seems having some issue.
Even no changes happened affecting coverage, it's degrading.

netil:
ping @sculove

sculove:
LGTM

Issue #169, opened 2019-04-11
When called in the order of dragging => restore method => resize method > dragging, the second dragging will not be moved into place.


daybrush:
## Description

The reason is that hold event occurs according to moveDistance status when hold in Axes. And initialize moveDistance during release.

If a restore occurs during a drag, the release event does not occur, but the moveDistance is not initialized, so the second drag does not generate a hold event, resulting in strange behavior.

PR #170, opened 2019-04-11
fix(eventHandler): fix changeHandler for stop

younkue:
## Issue

#169

### DIFF
--- a/src/eventHandler.js
+++ b/src/eventHandler.js
@@ -99,12 +99,12 @@ export default superclass => class extends superclass {
 			e.stop();
 			// release, animationEnd
 			touch.distance = 0;
-			touch.direction = direction || touch.direction;
+			touch.direction = null;
 			touch.destPos = 0;
 			this._adjustContainerCss("end");
 			touch.holding = false;
 			this._setPointerEvents();
-			conf.touch.isTrusted = false;
+			touch.isTrusted = false;
 			this._setPhaseValue("end");
 		}
 	}
--- a/src/eventHandler.js
+++ b/src/eventHandler.js
@@ -97,6 +97,15 @@ export default superclass => class extends superclass {
 			this._setTranslate([-pos, 0]);
 		} else {
 			e.stop();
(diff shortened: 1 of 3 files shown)


src/eventHandler.js
```diff
@@ -93,7 +93,20 @@ export default superclass => class extends superclass {
...
+		} else {
+			e.stop();
+			// release, animationEnd
+			touch.distance = 0;
+			touch.direction = direction || touch.direction;
+			touch.destPos = 0;
+			this._adjustContainerCss("end");
+			touch.holding = false;
+			this._setPointerEvents();
+			conf.touch.isTrusted = false;
```
WoodNeck: I think you can use just `touch` here.

src/eventHandler.js
```diff
@@ -93,7 +93,20 @@ export default superclass => class extends superclass {
...
 		);
 
-		(eventRes || eventRes === null) && this._setTranslate([-pos, 0]);
+		if (eventRes || eventRes === null) {
+			this._setTranslate([-pos, 0]);
+		} else {
+			e.stop();
+			// release, animationEnd
+			touch.distance = 0;
+			touch.direction = direction || touch.direction;
```
WoodNeck: Why we should maintain direction here? Isn't it like reset everything kind of phase?

src/eventHandler.js
```diff
@@ -93,7 +93,20 @@ export default superclass => class extends superclass {
...
+			e.stop();
+			// release, animationEnd
+			touch.distance = 0;
+			touch.direction = null;
+			touch.destPos = 0;
+			this._adjustContainerCss("end");
+			touch.holding = false;
+			this._setPointerEvents();
+			touch.isTrusted = false;
+			this._setPhaseValue("end");
```
jongmoon: @younkue How about grouping code into 2 groups. One group for `touch`, the other group is to call `method`?

jongmoon: @younkue Are these methods affected by order?

PR #176, opened 2019-04-23
fix: recalculate target panel when interrupted

### LINKED ISSUE #171: Target panel isn't recalculated if interrupted in circular mode
## Description
Version: Flicking 3.0.0

b11f1e10-40fd-11e9-8535-aa61c3fec8b5

1. Circular.
2. Animating to further than first cloned panel.
3. Interrrupted while moving.
Then, target panel isn't recalculated, so it will move further than expected.

### LINKED ISSUE #174: snap option is not from release position, but start position
## Description
`snap` option is applied from `currentPanel`, but should be applied on `nearestPanel`.

Test any flicking instance with `snap` option applied.
Drag more then one panel, then release with speed to test snap.

WoodNeck:
## Issue
Fixes #171 
It also resolves #174 

## Details
- This changes internal interface from FlickingPanel to real Panel's instance
- #171 is now fixed.
- #174 is also handled on this PR.

### DIFF
--- a/src/components/Panel.ts
+++ b/src/components/Panel.ts
@@ -167,7 +167,7 @@ class Panel implements FlickingPanel {
     const newPosition = currentPosition - prevPanelSize - options.gap;
 
     let prevPanel = prevSibling;
-    if (prevPanelPosition >= currentPosition) {
+    if (prevPanelPosition !== newPosition) {
       prevPanel = prevSibling.clone(prevSibling.getCloneIndex(), true);
       prevPanel.setPosition(newPosition, true);
     }
@@ -202,7 +202,7 @@ class Panel implements FlickingPanel {
     const newPosition = currentPosition + this.getSize() + options.gap;
 
     let nextPanel = nextSibling;
-    if (nextPanelPosition <= currentPosition) {
+    if (nextPanelPosition !== newPosition) {
       nextPanel = nextSibling.clone(nextSibling.getCloneIndex(), true);
       nextPanel.setPosition(newPosition, true);
     }
--- a/src/components/Panel.ts
+++ b/src/components/Panel.ts
@@ -133,7 +133,6 @@ class Panel implements FlickingPanel {
       ? ""
       : EVENTS.CHANGE;
 
-    console.log(duration);
     viewport.moveTo(this, eventType, null, duration);
   }
 
--- a/src/components/Panel.ts
+++ b/src/components/Panel.ts
@@ -1,15 +1,18 @@
-import { OriginalStyle, FlickingOptions } from "../types";
-import { DEFAULT_PANEL_CSS } from "../consts";
-import { addClass, applyCSS, parseArithmeticExpression } from "../utils";
+import Viewport from "./Viewport";
+import { OriginalStyle, FlickingPanel, ElementLike } from "../types";
+import { DEFAULT_PANEL_CSS, EVENTS } from "../consts";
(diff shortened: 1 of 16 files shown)


src/components/Panel.ts
```diff
@@ -74,6 +78,163 @@ class Panel {
...
+
+    const newPosition = currentPosition - prevPanelSize - options.gap;
+
+    let prevPanel = prevSibling;
+    if (prevPanelPosition >= currentPosition) {
+      prevPanel = prevSibling.clone(prevSibling.getCloneIndex(), true);
+      prevPanel.setPosition(newPosition, true);
+    }
+
+    return prevPanel;
```
daybrush: @WoodNeck 
It's a special case.
In circular mode, if enough virtual clones are created through prev() calls, then next(), the clones are not created.

src/components/Panel.ts
```diff
@@ -74,6 +78,163 @@ class Panel {
...
+    }
+
+    const newPosition = currentPosition + this.getSize() + options.gap;
+
+    let nextPanel = nextSibling;
+    if (nextPanelPosition <= currentPosition) {
+      nextPanel = nextSibling.clone(nextSibling.getCloneIndex(), true);
+      nextPanel.setPosition(newPosition, true);
+    }
+
```
daybrush: @WoodNeck 
The same is true if the opposite is done, as in the above comments.

src/components/Panel.ts
WoodNeck: Okay, I forgot to consider that one. I'll push fixing commit for it.

WoodNeck: I couldn't reproduce condition you mentioned. Can you give me some more details?

WoodNeck: Okay, now I got what's the problem.
I'll push fixing commit for it.

src/states/AnimatingState.ts
```diff
@@ -1,20 +1,38 @@
...
+    const scrollAreaSize = viewport.getScrollAreaSize();
+    const loopCount = Math.floor((this.lastPosition + this.delta - scrollArea.prev) / scrollAreaSize);
+
+    let targetPanel = this.targetPanel;
+    if (loopCount !== 0 && targetPanel) {
+      const cloneCount = viewport.panelManager.getCloneCount();
+      const originalTargetPosition = targetPanel.getPosition();
+
+      // cloneIndex is from -1 to cloneCount - 1
+      const newCloneIndex = circulate(targetPanel.getCloneIndex() - loopCount, -1, cloneCount - 1);
```
daybrush: @WoodNeck 
Is there a chance that newCloneIndex will get -1?

WoodNeck: As written in comment, cloneIndex is from -1 to cloneCount - 1
It's because original panel's cloneIndex is -1, and clone's cloneIndex starts from 0.
I've wrote code in that way as it's more convenient to manage clones like that.

So, newCloneIndex can be -1, if virtual panel is pointing original panel.

src/components/Panel.ts
```diff
@@ -74,6 +80,170 @@ class Panel {
...
+    if (hangerPosition === anchorPosition || !currentPanel) {
+      return;
+    }
+
+    const currentPosition = currentPanel.getPosition();
+    const eventType = currentPosition === this.getPosition()
+      ? ""
+      : EVENTS.CHANGE;
+
+    console.log(duration);
```
jongmoon: log remained.

WoodNeck: That file's outdated, and fixed on later commit.

PR #179, opened 2019-04-25
fix: resizing in freescroll

### LINKED ISSUE #177: When resize is called in "freeScroll" move type, panel is aligned.
## Description
3e2cc1cc-6084-11e9-9dfe-cc73e738b0c4
> `autoResize` is enabled

1. Init Flicking instance in "freeScroll" move type.
2. Call resize() when panel is not aligned.

### LINKED ISSUE #178: Changing DOM element's size in FlickingPanel's update() method breaks layout.
## Description
In FlickingPanel's `update()` method, any callback that changes panel's size is not applied automatically and breaks layout.

```js
panel.update(el => {
  // Do anything that changes panel's size, like changing CSS, innerHTML...
});

### PR DESCRIPTION
**Pull request by WoodNeck**
## Issue
This fixes #177 
This also fixes #178 

## Details
This PR fixes both #177 , #178
Also it adds some missing tests.

### DIFF
--- a/src/components/Viewport.ts
+++ b/src/components/Viewport.ts
@@ -376,11 +376,12 @@ export default class Viewport {
     panelManager.replace(index, panels);
 
     const currentPanel = this.currentPanel;
-    if (!currentPanel) {
+    const wasEmpty = !currentPanel;
+    if (wasEmpty) {
       this.currentPanel = panels[0];
-    } else if (isBetween(currentPanel.getIndex(), index, index + panels.length - 1)) {
+    } else if (isBetween(currentPanel!.getIndex(), index, index + panels.length - 1)) {
       // Current panel is replaced
-      this.currentPanel = panelManager.get(currentPanel.getIndex());
+      this.currentPanel = panelManager.get(currentPanel!.getIndex());
     }
 
     // Update checked indexes in infinite mode
@@ -394,6 +395,11 @@ export default class Viewport {
     });
 
     this.resize();
+
+    const isFreeScroll = (this.options.moveType as MoveTypeObjectOption).type === "freeScroll";
+    if (isFreeScroll && wasEmpty) {
+      this.moveTo(this.currentPanel!, "", null, 0);
+    }
 
     return panels;
   }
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: src/components/Viewport.ts [hunk 1]
```diff
@@ -844,7 +850,7 @@ export default class Viewport {
     // Update panel position && fit to wrapper
     let nextPanelPos = firstPanel.getPosition();
     let maintainingPanel: Panel = firstPanel;
-    if ((currentState.holding || currentState.playing) && nearestPanel) {
+    if (nearestPanel) {
```

What kind of conditions is this?

test/unit/methods.spec.ts
```diff
@@ -740,6 +741,28 @@ describe("Methods call", () => {
...
+      const prevCameraPosition = viewport.getCameraPosition();
+
+      // When
+      flicking.resize();
+
+      // Then
+      const afterCameraPosition = viewport.getCameraPosition();
+      expect(prevCameraPosition).equals(afterCameraPosition);
+    });
   });
```
daybrush: @WoodNeck 
Wouldn't it be nice to have a test when the pos goes out of range?

src/components/Viewport.ts
```diff
@@ -395,6 +396,11 @@ export default class Viewport {
 
     this.resize();
 
+    const isFreeScroll = (this.options.moveType as MoveTypeObjectOption).type === "freeScroll";
+    if (isFreeScroll && wasEmpty) {
+      this.moveTo(this.currentPanel!, "", null, 0);
```
jongmoon: @WoodNeck Is above code related with `resizing`?

src/components/Viewport.ts
WoodNeck: This was intended to update panel positions correctly only when in holding/animating state.
But in free scroll mode, it was also needed that it should be updated same in the idle state.

src/components/Viewport.ts
WoodNeck: Nope, it is related with adding new panel when there were no panels in free scroll mode.
As free scroll maintains camera position while in resizing, camera position usually is wrong after adding new panel.
So, I've manually modified it for its circumstance.

test/unit/methods.spec.ts
WoodNeck: That can be temporary fix.
To fix it correctly, we should modify Axes first so we can animating position as our own.
I think we should be able to change camera position while resizing first.

PR #183, opened 2019-04-26
fix: current panel is null

### LINKED ISSUE #180: current panel can be null after calling setStauts
## Description
`getCurrentPanel` can return null after calling `setStatus`, even though there's panel exists.
스크린샷 2019-04-26 오전 10 56 09

1. Initiate flicking without any panel.
2. Add panel at index 2, by calling `replace(2, panel)`
3. call `setStatus()` with `getStatus()`'s return value

WoodNeck:
## Issue
#180 

## Details
- This fixes #180

### DIFF
--- a/src/components/Viewport.ts
+++ b/src/components/Viewport.ts
@@ -475,6 +475,7 @@ export default class Viewport {
 
   public restore(status: FlickingStatus): void {
     const panels = status.panels;
+    const defaultIndex = this.options.defaultIndex;
     const cameraElement = this.cameraElement;
     const panelManager = this.panelManager;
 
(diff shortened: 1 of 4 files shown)


src/components/Viewport.ts
```diff
@@ -483,13 +484,18 @@ export default class Viewport {
...
 
     this.createPanels();
-    this.currentPanel = panelManager.get(status.index);
 
     // Reset panel index
     panelManager.originalPanels().forEach((panel, idx) => {
       panel.setIndex(panels[idx].index);
     });
 
+    this.currentPanel = panelManager.has(status.index)
```
How about `panelManager.get(status.index) || panelManager.get(defaultIndex) || panelManger.firstPanel()`?

WoodNeck: That looks more great, I'll fix it.

PR #186, opened 2019-04-26
fix: calling stop() to stop event immediately

### LINKED ISSUE #184: Calling stop() won't stop events immediately
## Description
Some events, like `holdStart`, `moveStart` won't be stopped immediately after calling event's `stop()` method.

1. Call stop() on `holdStart` event.
2. Another holdStart event can be triggered after `duration` given in FlickingOptions

### LINKED ISSUE #185: No needPanel event is triggered when there're no panels
## Description
If no panel exists, `needPanel` event won't be triggered

Init flicking instance with no panels, then call resize().

WoodNeck:
## Issue
#184, #185

## Details
- This fixes #184 
- This also fixes #185

### DIFF
--- a/src/states/DisabledState.ts
+++ b/src/states/DisabledState.ts
@@ -11,6 +11,12 @@ class DisabledState extends State {
     transitTo(STATE_TYPE.IDLE);
   }
 
+  public onChange(e: any, { transitTo }: FlickingContext): void {
+    // Can stop Axes's change event
+    e.stop();
+    transitTo(STATE_TYPE.IDLE);
+  }
+
   public onRelease(e: any, { transitTo }: FlickingContext): void {
     // This is needed when stopped hold start event
     if (e.delta.flick === 0) {
--- a/src/states/DisabledState.ts
+++ b/src/states/DisabledState.ts
@@ -11,9 +11,12 @@ class DisabledState extends State {
     transitTo(STATE_TYPE.IDLE);
   }
 
-  public onChange(e: any, { transitTo }: FlickingContext): void {
+  public onChange(e: any, { viewport, transitTo }: FlickingContext): void {
     // Can stop Axes's change event
     e.stop();
+
+    // Should update axes position as it's already changed at this moment
+    viewport.updateAxesPosition(viewport.getCameraPosition());
     transitTo(STATE_TYPE.IDLE);
   }
 
--- a/test/unit/events.spec.ts
+++ b/test/unit/events.spec.ts
@@ -174,7 +174,7 @@ describe("Events", () => {
     const testingEvents = [
       EVENTS.MOVE_START,
       EVENTS.MOVE,
-    ]
+    ];
 
     testingEvents.forEach(event => {
       it(`${event} can be stopped almost immediately`, () => {
--- a/test/unit/events.spec.ts
+++ b/test/unit/events.spec.ts
@@ -145,14 +145,16 @@ describe("Events", () => {
     });
     afterEach(() => {
       timer.restore();
-    })
+    });
 
     it("holdStart can be stopped almost immediately", () => {
       // Given
       let holdCount = 0;
+      const isPlayingInHoldStart: boolean[] = [];
+
       const flicking = flickingInfo.instance;
       flicking.on(EVENTS.HOLD_START, e => {
-        expect(flicking.isPlaying()).to.be.false;
+        isPlayingInHoldStart.push(flicking.isPlaying());
         holdCount += 1;
         e.stop();
       });
@@ -165,6 +167,7 @@ describe("Events", () => {
       timer.tick(500);
 
       // Then
+      expect(isPlayingInHoldStart.every(val => val === false)).to.be.true;
       expect(holdCount).equals(2);
     });
 
--- a/test/unit/events.spec.ts
+++ b/test/unit/events.spec.ts
@@ -164,7 +164,7 @@ describe("Events", () => {
       timer.tick(50);
 
       simulate(flickingInfo.element, { deltaX: -70, duration: 30 });
-      timer.tick(500);
+      timer.tick(500); // Make sure animation ends after second simulate()
 
       // Then
       expect(isPlayingInHoldStart.every(val => val === false)).to.be.true;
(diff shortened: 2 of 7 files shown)


test/unit/events.spec.ts
```diff
@@ -135,6 +137,65 @@ describe("Events", () => {
...
+    afterEach(() => {
+      timer.restore();
+    })
+
+    it("holdStart can be stopped almost immediately", () => {
+      // Given
+      let holdCount = 0;
+      const flicking = flickingInfo.instance;
+      flicking.on(EVENTS.HOLD_START, e => {
+        expect(flicking.isPlaying()).to.be.false;
```
jongmoon: @WoodNeck how about caching the result and Check value after `// Then`?

test/unit/events.spec.ts
```diff
@@ -135,6 +137,65 @@ describe("Events", () => {
...
+        holdCount += 1;
+        e.stop();
+      });
+
+      // When
+      simulate(flickingInfo.element, { deltaX: -70, duration: 30 });
+      timer.tick(50);
+
+      simulate(flickingInfo.element, { deltaX: -70, duration: 30 });
+      timer.tick(500);
```
jongmoon: @WoodNeck Why tick count is different from above(50)?

test/unit/events.spec.ts
```diff
@@ -135,6 +137,65 @@ describe("Events", () => {
...
+          e.stop();
+          eventCount += 1;
+        });
+
+        // When
+        simulate(flickingInfo.element, { deltaX: -70, duration: 30 });
+        timer.tick(50);
+
+        simulate(flickingInfo.element, { deltaX: -70, duration: 30 });
+        timer.tick(500);
```
jongmoon: @WoodNeck Why tick count is different from above(50)?

test/unit/events.spec.ts
```diff
@@ -135,6 +137,65 @@ describe("Events", () => {
...
+    ]
+
+    testingEvents.forEach(event => {
+      it(`${event} can be stopped almost immediately`, () => {
+        // Given
+        let eventCount = 0;
+        const flicking = flickingInfo.instance;
+        flicking.on(event, e => {
+          e.stop();
+          eventCount += 1;
```
jongmoon: Is result of previous code different from result of current code when `EVENTS.MOVE_START`?

test/unit/methods.spec.ts
```diff
@@ -1114,10 +1114,10 @@ describe("Methods call", () => {
       // Given
       const givenLastIndex = 5;
       const flicking = flickingInfo.instance;
-      flicking.setLastIndex(givenLastIndex);
 
       // When
-      await simulate(flickingInfo.element, { deltaX: -1000, duration: 50 });
+      flicking.setLastIndex(-1);
```
jongmoon: @WoodNeck What is the purpose of 1) setting `-1` and 2) moving setLastIndex order (after simulate)?

test/unit/events.spec.ts
WoodNeck: Okay, that would be more reasonable.

test/unit/events.spec.ts
WoodNeck: Previously, it was impossible to trigger holdStart before virtual animating ends if holdStart is stopped.
To test it is resolved, I made test to trigger holdStart twice in a really short term between.
So test can be success only if animation ends between first `tick()`
Second one is just to make sure animation ends after second holdStart.
Actually, it's useless I think.

test/unit/events.spec.ts
WoodNeck: In `holdStart` testing code, there's more `expect` test for whether it's animating.
That test is useless for `moveStart` and `move`, as it's explicitly animating.
But, looks like I can merge both tests I think.

test/unit/methods.spec.ts
WoodNeck: As setLastIndex calls resize(), then checks `needPanel` events should be triggered.
Previous test couldn't be passed as needPanel is triggered while calling `setLastIndex`, not simulating.
So, I've managed to fix it to call `setLastIndex(-1)` first, which removes all panels.
Then I call `setLastIndex()` again to trigger `needPanel`.

src/states/DisabledState.ts
```diff
@@ -11,6 +11,12 @@ class DisabledState extends State {
     transitTo(STATE_TYPE.IDLE);
   }
 
+  public onChange(e: any, { transitTo }: FlickingContext): void {
+    // Can stop Axes's change event
+    e.stop();
+    transitTo(STATE_TYPE.IDLE);
```
daybrush: @WoodNeck 

Change event in axes occurs and if stop method is called in the move event in flicking, go to disable mode. Then, won't the change event in axes stop when it happens one more time?

test/unit/events.spec.ts
jongmoon: @WoodNeck It would be better to comment it like `make sure animation ends after second holdStart`.

test/unit/events.spec.ts
jongmoon: @WoodNeck I mean... that previous code is not holdStart(which is above), it is not fixed code. I think this previous code will be woking fine although it's not fixed.

test/unit/events.spec.ts
jongmoon: @WoodNeck I got it.

test/unit/methods.spec.ts
jongmoon: Thank you for your kindness.

src/states/DisabledState.ts
WoodNeck: Fixed!

test/unit/events.spec.ts
WoodNeck: That is,
it's not like something directly related to this commit but it's more like testing Axes is working correctly, which is like checking whether using correct version of Axes or so.

test/unit/events.spec.ts
WoodNeck: That is,
it's not like something directly related to this commit but it's more like testing Axes is working correctly, which is like checking whether using correct version of Axes or so.

test/unit/events.spec.ts
WoodNeck: That would be nice, I'll fix it right away.

Issue #187, opened 2019-04-26
On Android 4.4.2, it does not work when crossing boundaries in circular mode


daybrush:
## Description

TypeError, no method 'sign'

PR #188, opened 2019-04-26
fix: remove unsupported method

daybrush:
## Issue

#187 

#195

### DIFF
--- a/src/Flicking.ts
+++ b/src/Flicking.ts
@@ -668,7 +668,7 @@ class Flicking extends Component {
       if (options.circular && looped) {
         // Reached at max/min range of axes
         const scrollAreaSize = viewport.getScrollAreaSize();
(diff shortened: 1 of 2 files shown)


src/Flicking.ts
```diff
@@ -668,7 +668,7 @@ class Flicking extends Component {
       if (options.circular && looped) {
         // Reached at max/min range of axes
         const scrollAreaSize = viewport.getScrollAreaSize();
-        cameraChange = -Math.sign(cameraChange) * (scrollAreaSize - Math.abs(cameraChange));
+        cameraChange = -(cameraChange > 0 ? 1 : -1) * (scrollAreaSize - Math.abs(cameraChange));
```
jongmoon: @daybrush It's not import issue.but I suggest `cameraChange < 0 ? -1 : 1` because 0 could be treated as negative.

PR #202, opened 2019-05-09
feat: make user to use existing Viewport and Camera element

### LINKED ISSUE #198: If Viewport or Camera element exists, use it.
## Description
Originally, Flicking generates its own Viewport and Camera DOM element to use.
But, in some case like using in framework-based environment like React, Vue, Angular.
It is more natural to have Viewport and Camera element to be generated by framework. 

## Spec
If element with class named {user-prefix}-flick-viewport, and {user-prefix}-flick-camera exists, use them for Viewport and Camera element.

WoodNeck:
## Issue
#198

## Details
- This resolves #198 
  - It also can cope with `destroy()` calls.
- It includes some refactoring, and also adds some tests for new feature.

### DIFF
--- a/src/components/Viewport.ts
+++ b/src/components/Viewport.ts
@@ -5,9 +5,9 @@ import Panel from "./Panel";
 import PanelManager from "./PanelManager";
 import StateMachine from "./StateMachine";
 import MoveType from "../moves/MoveType";
-import { FlickingOptions, FlickingPanel, FlickingStatus, ElementLike, EventType, TriggerCallback, NeedPanelEvent, FlickingEvent, MoveTypeObjectOption } from "../types";
+import { FlickingOptions, FlickingPanel, FlickingStatus, ElementLike, EventType, TriggerCallback, NeedPanelEvent, FlickingEvent, MoveTypeObjectOption, OriginalStyle } from "../types";
 import { DEFAULT_VIEWPORT_CSS, DEFAULT_CAMERA_CSS, TRANSFORM, DEFAULT_OPTIONS, EVENTS, DIRECTION, STATE_TYPE, MOVE_TYPE } from "../consts";
-import { clamp, applyCSS, toArray, parseArithmeticExpression, isBetween, isArray, parseElement } from "../utils";
+import { clamp, applyCSS, toArray, parseArithmeticExpression, isBetween, isArray, parseElement, hasClass, restoreStyle } from "../utils";
 import Snap from "../moves/Snap";
 import FreeScroll from "../moves/FreeScroll";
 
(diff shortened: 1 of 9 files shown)


src/components/Viewport.ts
```diff
@@ -473,16 +482,35 @@ export default class Viewport {
...
   public destroy(): void {
+    const state = this.state;
+    const wrapper = this.flicking.getElement();
     const viewportElement = this.viewportElement;
-    const wrapper = viewportElement.parentElement;
+    const cameraElement = this.cameraElement;
+
+    const originalPanels = this.panelManager.originalPanels();
+
+    restoreStyle(viewportElement, state.originalViewportStyle);
```
Do you think the condition to call restoreStyle function also requires `isCameraGiven` and `isViewportGiven`?

src/components/Viewport.ts
```diff
@@ -700,16 +729,78 @@ export default class Viewport {
...
+    const viewportCandidate = wrapper.children[0] as HTMLElement;
+    const hasViewportElement = hasClass(viewportCandidate, `${classPrefix}-viewport`);
+
+    const viewportElement = hasViewportElement
+      ? viewportCandidate
+      : document.createElement("div");
+
+    const cameraCandidate = hasViewportElement
+      ? viewportElement.children[0] as HTMLElement
+      : wrapper.children[0] as HTMLElement;
```
Is there a camera under the wrapper?

src/components/Viewport.ts
WoodNeck: Well that would be more efficient, but I think that's still trivial.

src/components/Viewport.ts
WoodNeck: I've considered all 4 possibilities.
Which is,
1) No Viewport, No Camera element
2) Viewport element exists, while no camera element.
3) Camera element exists, while no viewport element.
4) Both viewport and camera element exist.

test/unit/initialize.spec.ts
```diff
@@ -555,7 +609,10 @@ describe("Initialization", () => {
             const endIndex = flicking.getIndex();
             const indexAtDestPos = Math.floor(destPos / panelWidth);
             // As all delta is above threshold, it should change panel at least once
-            const expectedIndex = Math.min(nearestPanel.getIndex() + snapCount, Math.max(1, indexAtDestPos));
+            const expectedIndex = destPos >= 100
+              ? Math.min(nearestPanel.getIndex() + snapCount, Math.max(1, indexAtDestPos))
+              // Sometimes, simulation won't return proper dest position for given delta.
```
jongmoon: I suggest to prepend `TODO:` to check later. I think this code may have potential problem.

WoodNeck: It's been changed at other PR(#203). That's quite more detailed, so please check it after I rebase it.

PR #203, opened 2019-05-10
feat: support external rendering

### LINKED ISSUE #199: Support frameworks
## Description
Frameworks, like React, Vue, Angular has functionality to render elements by its own system.
As Flicking internally add/removes DOM elements, Flicking can't currently deal with such frameworks.

## Spec
- Add `sync` method, it will receive info of which panels are added, and changed its position. Flicking will then modify info of panels it is holding.
- Add `renderExternal: boolean` option. It will prevent Flicking from modifying(add/remove) elements

### LINKED ISSUE #200: Add getCloneCount
## Description
Clones are basically same with original panels, so it wasn't needed to consider it.
But in frameworks, they need to know about such infos like how many clone sets are needed.
They use that info to modify clone datas, to actually clone elements.

## Spec
Add `getCloneCount()`, which indicates how many cloned panel sets should be exist.

### LINKED ISSUE #205: Removing all panels in needPanel event will emit error.
## Description
Removing all panels at needPanel event will trigger error.

```js
// This will trigger error
flicking.on("needPanel", flicking.remove(0, Infinity);
```

### LINKED ISSUE #206: needPanel event isn't triggered when there is no panels.
## Description
`needPanel` event doesn't triggered when there're no panels inside Flicking.

1. Init Flicking only with wrapper, giving `infinite` option.
2. Do some flicking movement manually, or by method(prev, next ...).

WoodNeck:
## Issue
#199, #200 

## Details
- This closes #199 
- It also closes #200

- This PR is to support external rendering to use in frameworks like React, Vue, Angular
- It introduces 3 new interfaces
  - `renderExternal` option
  - `sync()` method
  - `getCloneCount()` method
- It also adds some tests.
- It refactored manual testing files, as it was too long in one file.

## Addition
- This closes #205, and also closes #206


src/components/Panel.ts
```diff
@@ -257,7 +261,7 @@ class Panel implements FlickingPanel {
   }
 
   public getElement(): HTMLElement {
-    return this.element;
+    return this.element!;
```
jongmoon: @WoodNeck What returns if `this.element` is null?

src/Flicking.ts
```diff
@@ -540,6 +541,76 @@ class Flicking extends Component {
...
+    const isCircular = this.options.circular;
+
+    // Make sure that new "list" should include cloned elements
+    const originalPanelCount = (list.length / (panelManager.getCloneCount() + 1)) >> 0; // Make sure it's integer. Same with Math.floor, but faster
+    const shouldCloneCount = ((list.length / originalPanelCount) >> 0) - 1;
+
+    const originalPanels = panelManager.originalPanels();
+    const clonedPanels = panelManager.clonedPanels();
+
+    const newOriginalElements = list.slice(0, originalPanelCount);
```
jongmoon: @WoodNeck Isn't possible to use `list` as newOriginalElements?

src/components/Panel.ts
WoodNeck: I thought it won't happen, as `element` will be `null` only if `renderExternal` is true.
And if it is, I thought setElement() will be called immediately after resizing in `sync()`.
I also think that kind of code is unstable, do you know any bypass for it?

src/Flicking.ts
WoodNeck: @daybrush said that list will be filled with `Element`, not indexes.
So I think that would be okay.

src/Flicking.ts
```diff
@@ -540,6 +541,76 @@ class Flicking extends Component {
...
+    const { list, added, maintained } = diffInfo;
+
+    const viewport = this.viewport;
+    const panelManager = viewport.panelManager;
+    const indexRange = panelManager.getRange();
+    const isCircular = this.options.circular;
+
+    // Make sure that new "list" should include cloned elements
+    const originalPanelCount = (list.length / (panelManager.getCloneCount() + 1)) >> 0; // Make sure it's integer. Same with Math.floor, but faster
+    const shouldCloneCount = ((list.length / originalPanelCount) >> 0) - 1;
```
jongmoon: @WoodNeck What's the meaning of `should` in `shouldCloneCount`?

src/Flicking.ts
```diff
@@ -540,6 +541,76 @@ class Flicking extends Component {
...
+
+    const originalPanels = panelManager.originalPanels();
```
jongmoon: @WoodNeck Does `originalPanels` include clones?

src/Flicking.ts
WoodNeck: It can interpreted to `given`.
As `list` is given, we should set cloneCount to that value.

src/Flicking.ts
WoodNeck: originalPanels doesn't include clones.
- originalPanels: Only originals
- clonedPanels: Only clones
- allPanels: Both of them

src/Flicking.ts
```diff
@@ -540,6 +541,76 @@ class Flicking extends Component {
...
+      newPanels[afterIdx] = originalPanels[beforeIdx];
+    });
+
+    originalAdded.forEach(addIndex => {
+      newPanels[addIndex] = new Panel(newOriginalElements[addIndex], addIndex, viewport);
+    });
+
+    if (isCircular) {
+      counter(shouldCloneCount).forEach(groupIndex => {
+        const originalCloneGroup = clonedPanels[groupIndex];
```
jongmoon: @WoodNeck How about to rename from `original` to `prev`?

src/Flicking.ts
jongmoon: How about `newCloneCount`?

src/Flicking.ts
WoodNeck: That would be more reasonable.

src/Flicking.ts
WoodNeck: That can be more nice, I'll fix it in next commit.

test/unit/initialize.spec.ts
```diff
@@ -555,7 +556,9 @@ describe("Initialization", () => {
             const endIndex = flicking.getIndex();
             const indexAtDestPos = Math.floor(destPos / panelWidth);
             // As all delta is above threshold, it should change panel at least once
-            const expectedIndex = Math.min(nearestPanel.getIndex() + snapCount, Math.max(1, indexAtDestPos));
+            const expectedIndex = eventDelta >= 50
```
jongmoon: @WoodNeck It would be better to comment for this.

WoodNeck: That's right, I forgot to add one. Thanks!

src/components/Panel.ts
```diff
@@ -1,62 +1,61 @@
...
       size: 0,
-      clonedPanels: [],
       isClone: false,
+      isVirtual: false,
       cloneIndex: -1,
       originalStyle: {
-        className: element.getAttribute("class") || null,
-        style: element.getAttribute("style") || null,
+        className: element ? element.getAttribute("class") : null,
+        style: element ? element.getAttribute("style") : null,
```
jongmoon: @WoodNeck Could element be null?

src/components/Panel.ts
```diff
@@ -1,62 +1,61 @@
...
     // Index of cloned panel, zero-based integer(original: -1, cloned: [0, 1, 2, ...])
     // if cloneIndex is 0, that means it's first cloned panel of original panel
     cloneIndex: number;
     originalStyle: OriginalStyle;
-    clonedPanels: Panel[];
     cachedBbox: ClientRect | null;
   };
+  private element: HTMLElement;
   private original?: Panel;
+  private clonedPanels: Panel[];
```
jongmoon: @WoodNeck Why does clonedPanels come out from state?

src/components/Panel.ts
WoodNeck: Oh, I forgot to get rid of that thing. I think it should be
```js
className: element.getAttribute("class")
```
I'll fix it right away. Thanks.

src/components/Panel.ts
WoodNeck: I thought it was inappropriate to store clonedPanels in state, as I thought `state` is to store some values to be save / restored in getStatus() / setStatus() (saving panel infos is not implemented though)

src/Flicking.ts
```diff
@@ -531,6 +546,90 @@ class Flicking extends Component {
...
+          const newPanel = newPanels[addIndex];
+
+          newCloneGroup[addIndex] = newPanel.cloneExternal(groupIndex, newClonedElements[cloneGroupOffset + addIndex]);
+        });
+      });
+    }
+
+    // Replace current info of panels this holds
+    added.forEach(index => { viewport.updateCheckedIndexes({ min: index, max: index }); });
+    removed.forEach(index => { viewport.updateCheckedIndexes({ min: index - 1, max: index + 1 }); });
```
jongmoon: @WoodNeck Is order important? eg. `added -> removed` or `removed ->added`

src/Flicking.ts
```diff
@@ -531,6 +546,90 @@ class Flicking extends Component {
...
+
+    const checkedIndexes = viewport.getCheckedIndexes();
+    checkedIndexes.forEach(([min, max], idx) => {
+      const pushedIndex = added.filter(index => index < min).length
+        - removed.filter(index => index < min).length;
+      checkedIndexes.splice(idx, 1, [min + pushedIndex, max + pushedIndex]);
```
jongmoon: @WoodNeck Is this what you said on offline?

src/Flicking.ts
WoodNeck: I think not. Only order of `maintained` is important.

src/Flicking.ts
WoodNeck: Nope, it's Viewport's `updateCheckedIndexes()` method.

src/components/Viewport.ts
```diff
@@ -445,6 +447,19 @@ export default class Viewport {
...
+
+    // Update checked indexes in infinite mode
+    if (deleteCount > 0) {
+      state.checkedIndexes.forEach((indexes, idx) => {
+        const [min, max] = indexes;
+        // Can fill part of indexes in range
+        if (isBetween(index, min, max)) {
+          // Remove checked index from list
+          state.checkedIndexes.splice(idx, 1);
+        }
```

There are a lot of splice codes in forEach. I'm afraid we'll have an index problem, so please check.

src/Flicking.ts
```diff
@@ -535,6 +550,94 @@ class Flicking extends Component {
...
+   * @param {number[]} [diffInfo.added] - Index array of DOM elements added to `list`.<ko>`list`에서 추가된 DOM 요소들의 인덱스 배열.</ko>
+   * @param {number[]} [diffInfo.removed] - Index array of DOM elements removed from previous element list.<ko>이전 리스트에서 제거된 DOM 요소들의 인덱스 배열.</ko>
+   */
+  public sync(diffInfo: {
+    list: HTMLElement[],
+    maintained: number[][],
+    added: number[],
+    removed: number[],
+  }) {
+    const { list, maintained, added, removed } = diffInfo;
```
daybrush: @WoodNeck 

How about using `changed`?

if (!added.length && !removed.length && !changed.length)  {
    return this;
}
```

File: src/Flicking.ts [hunk 13]
```diff
@@ -618,8 +620,18 @@ class Flicking extends Component {
...
 
     // Replace current info of panels this holds
+    added.forEach(index => { viewport.updateCheckedIndexes({ min: index, max: index }); });
+    removed.forEach(index => { viewport.updateCheckedIndexes({ min: index - 1, max: index + 1 }); });
+
+    const checkedIndexes = viewport.getCheckedIndexes();
+    checkedIndexes.forEach(([min, max], idx) => {
+      const pushedIndex = added.filter(index => index < min).length
+        - removed.filter(index => index < min).length;
+      checkedIndexes.splice(idx, 1, [min + pushedIndex, max + pushedIndex]);
```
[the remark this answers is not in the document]
WoodNeck: This code just updates value at `idx`, so it won't make any problem I guess.

src/components/Panel.ts
```diff
@@ -336,15 +335,18 @@ class Panel implements FlickingPanel {
 
   public clone(cloneIndex: number, virtual: boolean = false): Panel {
```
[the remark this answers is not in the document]
WoodNeck: That would be nice

src/components/Viewport.ts
WoodNeck: Sure, that must be fixed.

src/Flicking.ts
WoodNeck: I'll do it right away~

src/components/Viewport.ts
WoodNeck: I've checked it but this looks outdated, already fixed in previous commit.

src/Flicking.ts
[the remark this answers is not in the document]
daybrush: `&&` would be better than `+` to minimize access to length.

Issue #217, opened 2019-05-22
Add getLastIndex


WoodNeck:
## Description
`getLastIndex()` should be made, as we have to slice child nodes with that value when we're doing external rendering.

PR #218, opened 2019-05-22
feat: add getLastIndex

WoodNeck:
## Issue
#217

## Details
- This adds new method, `getLastIndex()`, to cope with external rendering.
- Did a few test complexity refactoring.

### DIFF
--- a/src/Flicking.ts
+++ b/src/Flicking.ts
@@ -123,13 +123,11 @@ class Flicking extends Component {
     const currentPanel = this.getCurrentPanel();
     const currentState = this.viewport.stateMachine.getState();
 
-    if (!currentPanel || currentState.type !== STATE_TYPE.IDLE) {
-      return this;
-    }
-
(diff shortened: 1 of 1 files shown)


src/Flicking.ts
```diff
@@ -123,13 +123,11 @@ class Flicking extends Component {
     const currentPanel = this.getCurrentPanel();
     const currentState = this.viewport.stateMachine.getState();
 
-    if (!currentPanel || currentState.type !== STATE_TYPE.IDLE) {
-      return this;
```
jongmoon: @WoodNeck I wonder why do you think it makes complexity?

WoodNeck: I'm currently using [CodeMetrics](https://marketplace.visualstudio.com/items?itemName=kisstkondoros.vscode-codemetrics) for practicing purpose.
They regard `return` phrase makes more complexity, so I've refactored it.

Issue #220, opened 2019-05-24
Errors occur when used in node.js(ssr) environments such as react, angular, and vue.


daybrush:
## Description

```
ReferenceError: document is not defined
    at checkTranslateSupport (/Users/user/Desktop/test/next-learn-demo/5-clean-urls-ssr/node_modules/@egjs/flicking/dist/flicking.js:99:28)
    at /Users/user/Desktop/test/next-learn-demo/5-clean-urls-ssr/node_modules/@egjs/flicking/dist/flicking.js:446:21
    at /Users/user/Desktop/test/next-learn-demo/5-clean-urls-ssr/node_modules/@egjs/flicking/dist/flicking.js:10:85
    at Object.<anonymous> (/Users/user/Desktop/test/next-learn-demo/5-clean-urls-ssr/node_modules/@egjs/flicking/dist/flicking.js:13:2)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:22:18)
    at Object.<anonymous> (/Users/user/Desktop/test/next-learn-demo/5-clean-urls-ssr/node_modules/@egjs/react-flicking/dist/flicking.cjs.js:11:22)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)

```

PR #221, opened 2019-05-24
fix: fix bug for node.js

daybrush:
## Issue

#220

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -50,10 +50,7 @@ export let checkTranslateSupport = () => {
   };
 
   if (typeof document === "undefined") {
-    return {
-      name: "",
-      has3d: false,
-    };
+    return;
   }
   const supportedStyle = document.documentElement.style;
   let transformName = "";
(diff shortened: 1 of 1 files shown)


src/utils.ts
```diff
@@ -49,6 +49,12 @@ export let checkTranslateSupport = () => {
     transform: "transform",
   };
 
+  if (typeof document === "undefined") {
+    return {
+      name: "",
+      has3d: false,
+    };
```
jongmoon: @daybrush Why didn't you return null instead of obj?

WoodNeck:
Does React have some feature like this?
https://ko.nuxtjs.org/guide/plugins/#%ED%81%B4%EB%9D%BC%EC%9D%B4%EC%96%B8%ED%8A%B8%EC%97%90%EC%84%9C%EB%A7%8C-%EC%82%AC%EC%9A%A9%ED%95%98%EA%B8%B0

By enabling this, it won't make any problem with Vue SSR(Nuxt.js)

daybrush:
@WoodNeck I don't know in detail, but I've seen using Dynamic Imports if it's not available or used by clients.

The plug-in above seems to define that certain parts will not be used with ssr.

PR #224, opened 2019-05-27
fix: change event is triggered on bound area

### LINKED ISSUE #208: change event is triggered at bound area
## Description
`change` event is triggered at bound area when Flicking is initialized with option
`bounce: 0, bound: true, moveType: freeScroll`

1. Initialize Flicking with above options.
2. Scroll to bound area

WoodNeck:
## Issue
#208

## Details
This fixes #208

### DIFF
--- a/src/moves/FreeScroll.ts
+++ b/src/moves/FreeScroll.ts
@@ -6,7 +6,7 @@
 import Snap from "./Snap";
 import { MOVE_TYPE, EVENTS } from "../consts";
 import { MoveTypeContext, DestinationInfo } from "../types";
-import { circulate } from "../utils";
+import { circulate, isBetween } from "../utils";
 
 class FreeScroll extends Snap {
   protected readonly type: string = MOVE_TYPE.FREE_SCROLL;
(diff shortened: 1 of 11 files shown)


src/moves/FreeScroll.ts
```diff
@@ -65,13 +89,13 @@ class FreeScroll extends Snap {
     // Ref #191(https://github.com/naver/egjs-flicking/issues/191)
     const lastHangerPosition = lastPosition + viewport.getRelativeHangerPosition();
 
+    const scrollAreaSize = viewport.getScrollAreaSize();
     let minimumDistanceToChange = isNextDirection
       ? currentPanelPosition + currentPanel.getSize() - lastHangerPosition + halfGap
       : lastHangerPosition - currentPanelPosition + halfGap;
+    minimumDistanceToChange = Math.abs(minimumDistanceToChange % scrollAreaSize);
```
daybrush: @WoodNeck Why did you get the rest of it?

WoodNeck: In scroll boundary area in circular mode, `minimumDistanceToChange` can be calculated from cloned panel 0 to original panel 0 or vice versa.
In that case, distance from original panel 0 to cloned panel 0 is added to `minimumDistanceToChange` and `scrollAreaSize` is it.
So by taking modulus, we can calculate correct `minimumDistanceToChange`

PR #225, opened 2019-05-27
fix: every panel to be initialized with CSS "left"

### LINKED ISSUE #213: CSS "left" value of first panel is not set when initialized
## Description
First panel's CSS `left` not exists after initialized. Not even `0px` is set.

## Cause
As Panel's default position is 0, and as Panel won't update position value if it's same value as currently holding, first panel's `left` won't be set.

WoodNeck:
## Issue
#213

## Details
This fixes #213

### DIFF
--- a/src/components/Viewport.ts
+++ b/src/components/Viewport.ts
@@ -1065,12 +1065,9 @@ export default class Viewport {
 
     panels.forEach(panel => {
       const newPosition = nextPanelPos;
-      const currentPosition = panel.getPosition();
       const panelSize = panel.getSize();
 
-      if (currentPosition !== newPosition) {
-        panel.setPosition(newPosition);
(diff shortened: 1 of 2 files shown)


src/components/Viewport.ts
```diff
@@ -710,7 +710,7 @@ export default class Viewport {
   }
 
   public setCurrentPanel(panel: Panel): void {
-    this.currentPanel = panel;
+    this.currentPanel = panel.getOriginalPanel();
```
jongmoon: @WoodNeck What is the purpose of this code? Doesn't it trigger more move?

WoodNeck: That's my mistake.
Looks like codes from other commit not omitted and included on this one.

PR #228, opened 2019-05-28
feat: relative re-positioning in freeScroll

### LINKED ISSUE #226: Rotating screen will make blank space in flicking area if free scroll is enabled
## Description
When free scroll option is enabled and rotating screen will create blank space 
image

With option
```js
{
  "moveType": "freeScroll",
  "bound": true,
  "bounce": 0,
  "autoResize": true,
}
```

1. While on landscape mode, move to last panel
2. Rotate to portrait mode.

### LINKED ISSUE #227: Relative camera positioning
## Description
To fix #226, I suggest to store one more info for camera.
That is, relative position of camera(hanger) above panel.
By doing that, skipping camera position update while free scroll or animating check can be disabled.
```js
// This can be removed
const isFreeScroll = (this.options.moveType as MoveTypeObjectOption).type === "freeScroll";

if (currentState.holding || currentState.playing || isFreeScroll) {
    return;
}
```

WoodNeck:
## Issue
#226, #227

## Details
This fixes #226
This resolves #227 

## BEFORE
before

## AFTER
after

### DIFF
--- a/src/components/Viewport.ts
+++ b/src/components/Viewport.ts
@@ -39,7 +39,7 @@ export default class Viewport {
   private state: {
     size: number;
     position: number;
-    relativePosition: number;
+    panelMaintainRatio: number;
     relativeHangerPosition: number;
     scrollArea: {
       prev: number;
@@ -68,7 +68,7 @@ export default class Viewport {
     this.state = {
       size: 0,
       position: 0,
-      relativePosition: 0,
+      panelMaintainRatio: 0,
       relativeHangerPosition: 0,
       scrollArea: {
         prev: 0,
@@ -184,9 +184,9 @@ export default class Viewport {
       const halfGap = options.gap / 2;
 
       // As panel's range is from panel position - half gap ~ panel pos + panel size + half gap
-      state.relativePosition = (hangerPosition - panelPosition + halfGap) / (panelSize +  2 * halfGap);
+      state.panelMaintainRatio = (hangerPosition - panelPosition + halfGap) / (panelSize +  2 * halfGap);
     } else {
-      state.relativePosition = 0;
+      state.panelMaintainRatio = 0;
     }
 
     this.checkNeedPanel(axesEvent);
@@ -1201,12 +1201,12 @@ export default class Viewport {
       return;
     }
 
-    let newPosition = 0;
+    let newPosition: number;
     if (isFreeScroll) {
       const nearestPanel = this.getNearestPanel();
 
       newPosition = nearestPanel
-        ? nearestPanel.getPosition() - halfGap + (nearestPanel.getSize() + 2 * halfGap) * state.relativePosition - relativeHangerPosition
+        ? nearestPanel.getPosition() - halfGap + (nearestPanel.getSize() + 2 * halfGap) * state.panelMaintainRatio - relativeHangerPosition
         : this.getCameraPosition();
     } else {
       newPosition = currentPanel
--- a/src/components/Viewport.ts
+++ b/src/components/Viewport.ts
@@ -39,6 +39,7 @@ export default class Viewport {
   private state: {
     size: number;
     position: number;
+    relativePosition: number;
     relativeHangerPosition: number;
     scrollArea: {
       prev: number;
@@ -67,6 +68,7 @@ export default class Viewport {
     this.state = {
       size: 0,
       position: 0,
+      relativePosition: 0,
       relativeHangerPosition: 0,
       scrollArea: {
         prev: 0,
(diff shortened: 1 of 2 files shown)


src/components/Viewport.ts
```diff
@@ -168,12 +170,25 @@ export default class Viewport {
...
+    // From 0(panel position) to 1(panel position + panel size)
+    // When it's on gap area value will be (val > 1 || val < 0)
+    if (nearestPanel) {
+      const hangerPosition = this.getHangerPosition();
+      const panelPosition = nearestPanel.getPosition();
+      const panelSize = nearestPanel.getSize();
+      const halfGap = options.gap / 2;
+
+      // As panel's range is from panel position - half gap ~ panel pos + panel size + half gap
+      state.relativePosition = (hangerPosition - panelPosition + halfGap) / (panelSize +  2 * halfGap);
```
The relativeHangerPosition, relativeAnchorPosition used for the state is known as just a position, but relativePosition here is the ratio of position. The name is vague.

And how about using getProgress?

WoodNeck: I agree name's very vague, but I don't like syntax of `getProgress`.
Putting same value twice looks very awful for me.
Like, `getProgress([hP, pP - hG, pP - hG, pP +pS + hG])`.
Also, it's a bit more efficient for a simple calculation like that.

src/components/Viewport.ts
```diff
@@ -1178,15 +1193,26 @@ export default class Viewport {
...
 
-    if (currentState.holding || currentState.playing || isFreeScroll) {
+    if (currentState.holding || currentState.playing) {
       return;
     }
 
-    let newPosition = currentPanel
-      ? currentPanel.getAnchorPosition() - state.relativeHangerPosition
-      : this.getCameraPosition();
+    let newPosition = 0;
```
jongmoon: 0 assign looks unnecessary~

src/components/Viewport.ts
daybrush: It meant using nearestPanel.getProgress. But I think the way it is now is better.

src/components/Viewport.ts
WoodNeck: I've fixed it on e6fb12c

src/components/Viewport.ts
```diff
@@ -168,12 +170,25 @@ export default class Viewport {
...
+    // From 0(panel position) to 1(panel position + panel size)
+    // When it's on gap area value will be (val > 1 || val < 0)
+    if (nearestPanel) {
+      const hangerPosition = this.getHangerPosition();
+      const panelPosition = nearestPanel.getPosition();
+      const panelSize = nearestPanel.getSize();
+      const halfGap = options.gap / 2;
+
+      // As panel's range is from panel position - half gap ~ panel pos + panel size + half gap
+      state.panelMaintainRatio = (hangerPosition - panelPosition + halfGap) / (panelSize +  2 * halfGap);
```
jongmoon: @WoodNeck How about to change from `hangerPosition` to `cameraPosition`?

src/components/Viewport.ts
```diff
@@ -1178,15 +1193,26 @@ export default class Viewport {
...
 
-    let newPosition = currentPanel
-      ? currentPanel.getAnchorPosition() - state.relativeHangerPosition
-      : this.getCameraPosition();
+    let newPosition: number;
+    if (isFreeScroll) {
+      const nearestPanel = this.getNearestPanel();
+
+      newPosition = nearestPanel
+        ? nearestPanel.getPosition() - halfGap + (nearestPanel.getSize() + 2 * halfGap) * state.panelMaintainRatio - relativeHangerPosition
```
jongmoon: Could relativePosition be omitted if `panelMaintainRatio` is calculated by `cameraPosition` instead of `hangerPosition`?

src/components/Viewport.ts
WoodNeck: It's meaningful to use hanger position, as where camera position actually affected by panel is hanger position, not camera position.

Anyway, it should be related with hanger position.
If using camera position instead, it would be like
```js
state.panelMaintainRatio = (cameraPosition - (panelPosition - relativeHangerPosition - halfGap)) / (panelSize + 2 * halfGap);
```

src/components/Viewport.ts
WoodNeck: I think not. It can be same.

```js
nearestPanel.getPosition() - relativeHangerPosition - halfGap + (nearestPanel.getSize() + 2 * halfGap) * state.panelMaintainRatio
```

src/components/Viewport.ts
jongmoon: @WoodNeck I got it! Thank you!

Issue #230, opened 2019-05-28
In freeScroll mode, duration also happens to be very short when the distance to move is short.


daybrush:
## Description

In freeScroll mode, duration also happens to be very short when the distance to move is short.

freeScroll is not using axesEvent duration.

PR #231, opened 2019-05-28
fix: fix freeScroll's duration

daybrush:
## Issue

#230

### DIFF
--- a/src/moves/FreeScroll.ts
+++ b/src/moves/FreeScroll.ts
@@ -27,6 +27,8 @@ class FreeScroll extends Snap {
     const delta = Math.abs(axesEvent.delta.flick + state.delta);
     if (delta > minimumDistanceToChange) {
       const destInfo = super.findSnappedPanel(ctx);
+
+      destInfo.duration = Math.max(axesEvent.duration, destInfo.duration);
       destInfo.destPos = destPos;
       destInfo.eventType = !options.circular && destInfo.panel === currentPanel
         ? ""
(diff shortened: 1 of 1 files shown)


src/moves/FreeScroll.ts
```diff
@@ -56,12 +58,12 @@ class FreeScroll extends Snap {
   }
 
   public findPanelWhenInterrupted(ctx: MoveTypeContext): DestinationInfo {
-    const { viewport } = ctx;
+    const { viewport, axesEvent } = ctx;
 
     return {
       panel: viewport.getNearestPanel()!,
       destPos: viewport.getCameraPosition(),
-      duration: viewport.options.duration,
```
WoodNeck: This duration can't be more than 0 as it's just stopping on camera position. It's my fault.
Could you please fix that for me?

src/moves/FreeScroll.ts
```diff
@@ -45,7 +47,7 @@ class FreeScroll extends Snap {
       return {
         panel: estimatedPanel,
         destPos,
-        duration: viewport.options.duration,
+        duration: Math.max(axesEvent.duration, viewport.options.duration),
```
jongmoon: @daybrush I think duration should follow only axesEvent.duration in freeScroll mode because viewport.option.duration is only applied when moving between panels.

src/moves/FreeScroll.ts
```diff
@@ -27,6 +27,8 @@ class FreeScroll extends Snap {
     const delta = Math.abs(axesEvent.delta.flick + state.delta);
     if (delta > minimumDistanceToChange) {
       const destInfo = super.findSnappedPanel(ctx);
+
+      destInfo.duration = Math.max(axesEvent.duration, destInfo.duration);
```
jongmoon: @daybrush I think `destInfo.duration = axesEvent.duration;`

src/moves/FreeScroll.ts
daybrush: @WoodNeck It was a fix.

PR #233, opened 2019-05-29
feat: add destroy()'s new option - "preserveUI"

### LINKED ISSUE #232: Option to preserve style when destroy() called
## Description
As current `destroy()` always removes all styles generated by Flicking, what we see is can be changed dramatically before/after `destroy()` is called.

To handle this, basically add an option for the destroy method, which enables preserving styles / elements.

WoodNeck:
## Issue
#232 

## Details
This resolves #232

### DIFF
--- a/src/Flicking.ts
+++ b/src/Flicking.ts
@@ -9,7 +9,7 @@ import Panel from "./components/Panel";
 
 import { merge, getProgress, parseElement, isString, counter } from "./utils";
 import { DEFAULT_OPTIONS, EVENTS, DIRECTION, AXES_EVENTS, STATE_TYPE, DEFAULT_MOVE_TYPE_OPTIONS } from "./consts";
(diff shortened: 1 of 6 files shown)


src/Flicking.ts
```diff
@@ -427,10 +427,14 @@ class Flicking extends Component {
    * flick.destroy();
    * console.log(flick.moveTo); // null
    */
-  public destroy(): void {
+  public destroy(option: Partial<DestroyOption> = {}): void {
```
daybrush: @WoodNeck 
Don't you think it's okay if you have one option and it's not an object?

WoodNeck: It's for flexibility.
Whenever we need to add more option for the `destroy()`, we just have to put more option in it.
This is because we might need more options for it, like should preserve only elements but not styles? or only styles? or so.

PR #238, opened 2019-06-05
feat: init ngx-flicking

jongmoon:
## Issue
Ref #199

# Limitation or Discussion
## `<ngx-flicking>` vs `<eg-flicking>`
  - [Angular Style Guide for Component Selector](https://angular.io/guide/styleguide#style-05-02)
    - kebab-case
    - Prevent name collision
    - Easy to identify

### `<eg-flicking>`
  - Example: Major product designates alphabetic characters representing the representative branding as prefix
    - [PRIMENG](https://www.primefaces.org/primeng): p-
    - [material components](https://material.angular.io/components/): mat-
    - [ng-zorro](https://ng.ant.design/components): nz-
    - [clarity](https://v1.clarity.design/): clr-
    - [angular-calendar](https://github.com/mattlewis92/angular-calendar): mwl-

### `<ngx-flicking>`
  - Example: In the case of ng-, ngx-, it is often used in a somewhat minor product
    - [ng-select](https://github.com/ng-select/ng-select): ng-
    - [ng2-table](https://github.com/valor-software/ng2-table/blob/master/demo/components/table/table-demo.html): ng-
    - [ngx-siema](https://github.com/lexzhukov/ngx-siema): ngx
    - [ngx-charts-chart](https://github.com/swimlane/ngx-charts): ngx-

### Event Handling
  - Handle events through handleEvents without handling events individually
  - Like Vue or React, [It can not register events to be processed by iteration](https://oss.navercorp.com/egjs/work/issues/355#issuecomment-3222335)

### lastIndex - impossible
  - Failed to access panel information injected by parents.
    - What alternatives should be provided?

### Can not figure out when flicking was fully updated
  - Same issue as Vue
  - How should I solve it?

### Can not be dynamically tagged
  - Branching to conditional statements is possible


packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
```diff
@@ -0,0 +1,128 @@
...
+  @Input() options: Partial<FlickingOptions> = {};
+  @Input() plugins: Plugin[] = [];
+
+  // Reference Flicking.EventType
+  @Output() handleEvents = new EventEmitter<Partial<FlickingEvent>>();
+  @ContentChild(TemplateRef) template: TemplateRef<any>;
+
+  @withFlickingMethods
+  private flicking?: NativeFlicking | null;
+  private pluginsDiffer: ListDiffer<Plugin> = new ListDiffer<Plugin>();
```
daybrush: @jongmoon 
I think it's better not to use private accessors in a consistent way.

packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  private bindEvents() {
+    const events = Object.keys(NativeFlicking.EVENTS)
+      .map(key => NativeFlicking.EVENTS[key]);
+
+    events.forEach(eventName => {
+      this.flicking.on(eventName, e => {
+        e.currentTarget = this;
+
+        this.handleEvents.emit(e); // type is camelCase
```
daybrush: @jongmoon I think it's better to call individually.

packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
```diff
@@ -0,0 +1,128 @@
...
+ * Copyright (c) 2015 NAVER Corp.
+ * egjs projects are licensed under the MIT license
+ */
+import NativeFlicking, { Plugin, FlickingOptions, withFlickingMethods, DEFAULT_OPTIONS, FlickingEvent } from '@egjs/flicking';
+// tslint:disable-next-line: max-line-length
+import { Component, OnInit, Input, AfterViewInit, ElementRef, OnChanges, Output, EventEmitter, OnDestroy, ContentChild, TemplateRef } from '@angular/core';
+import ListDiffer, { DiffResult } from '@egjs/list-differ';
+
+@Component({
+  selector: 'eg-flicking',
```
daybrush: @jongmoon Why don't you use ngx-flickering for the selector?

packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
jongmoon: @daybrush 
`classPrefix` and `cloneCount` are used in `html template`. So It need not to use private. But others is not used on others. I think It should be private if it is intended as public.

packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
jongmoon: @daybrush I researched angular component and found that **major product** use their namespace as a prefix. 

But I agree with you because product name is `ngx-flicking` and `eg-` would make some confusion.

packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
jongmoon: @daybrush OK. I agree. I've researched other library and they did it indivisually.

packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
daybrush: @jongmoon 
I thought the selector was `ngx-flicking` because the class name is `NgxFlicking`. Were the components you were looking for the same class name and selector name?

packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
```diff
@@ -121,7 +130,8 @@ export class NgxFlickingComponent implements OnInit, AfterViewInit, OnDestroy, O
       this.flicking.on(eventName, e => {
         e.currentTarget = this;
 
-        this.handleEvents.emit(e); // type is camelCase
+        // Style guide: Event - https://angular.io/guide/styleguide#dont-prefix-output-properties
+        this[eventName].emit(e);
```
daybrush: @jongmoon 

It will likely be necessary to deal with events that are added later.

jongmoon: @daybrush Right! 👍

packages/ngx-flicking/projects/ngx-flicking/package.json
```diff
@@ -1,12 +1,28 @@
 {
-  "name": "ngx-flicking",
-  "version": "0.1.0",
+  "name": "@egjs/ngx-flicking",
+  "version": "3.0.0-beta",
+  "description": "A module used to implement flicking interactions. With this module, you can make flicking gestures, which are ways to navigate left and right to move between panels arranged side by side.",
   "peerDependencies": {
     "@angular/common": "^7.2.0",
-    "@angular/core": "^7.2.0"
+    "@angular/core": "^7.2.15"
```
WoodNeck: Which Angular feature made this change?

packages/ngx-flicking/projects/ngx-flicking/package.json
```diff
@@ -1,12 +1,28 @@
...
   },
   "dependencies": {
     "@egjs/flicking": "^3.1.1",
     "@egjs/ngx-children-differ": "^0.1.0"
-  }
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/naver/egjs-flicking"
```
WoodNeck: I think we should use angular version README file instead.
That is, https://github.com/naver/egjs-flicking/tree/master/packages/ngx-flicking

packages/ngx-flicking/projects/ngx-flicking/package.json
```diff
@@ -1,12 +1,28 @@
...
+  },
+  "author": {
+    "name": "NAVER Corp."
+  },
+  "namespace": {
+    "eg": "eg"
+  },
+  "homepage": "https://github.com/naver/egjs-flicking",
```
WoodNeck: I've fixed homepage to indicate our github page in native flicking.
Could you please fix this to https://naver.github.io/egjs-flicking ?

packages/ngx-flicking/projects/ngx-flicking/package.json
jongmoon: @WoodNeck It is not intended change. But it would not be problem by following reason.
  1. It's patch (bug fix)
  2. It's peerDependency

PR #244, opened 2019-06-11
fix: clone count calculation

### LINKED ISSUE #241: Clone count is miscalculated
## Description
Flicking's clone count is miscalculated in circular mode.

https://codepen.io/woodneck/pen/jjNmaG

### LINKED ISSUE #243: Resize makes more clones
## Description
`resize()` makes more cloned panels at initialization.

Call flicking's resize() method at initialization.

WoodNeck:
## Issue
#241, #243 

## Details
This fixes #241
It also fixes #243

### DIFF
--- a/src/components/Viewport.ts
+++ b/src/components/Viewport.ts
@@ -1000,13 +1000,13 @@ export default class Viewport {
     }
 
     // Need one more set of clones on prev area of original panel 0
-    const needCloneOnPrev = panelAtLeftBoundary.getIndex() <= panelAtRightBoundary.getIndex()
-      && panelAtLeftBoundary.getIndex() !== 0;
+    const needCloneOnPrev = panelAtLeftBoundary.getIndex() !== 0
+      && panelAtLeftBoundary.getIndex() <= panelAtRightBoundary.getIndex();
 
     // Visible count of panel 0 on first screen
     const panel0OnFirstscreen = Math.ceil((relativeHangerPosition + firstPanel.getSize() - relativeAnchorPosition) / sumOriginalPanelSize)
       + Math.ceil((viewportSize - relativeHangerPosition + relativeAnchorPosition) / sumOriginalPanelSize)
-      - 1;
+      - 1; // duplication
 
     const cloneCount = panel0OnFirstscreen
       + (needCloneOnPrev ? 1 : 0);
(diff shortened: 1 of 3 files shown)


src/components/Viewport.ts
```diff
@@ -953,26 +953,67 @@ export default class Viewport {
...
+      sizeSum += panel.getSize() + gap;
+      if (sizeSum >= areaNext) {
+        panelAtRightBoundary = panel;
+        break;
+      }
+    }
+
+    // Need one more set of clones on prev area of original panel 0
+    const needCloneOnPrev = panelAtLeftBoundary.getIndex() <= panelAtRightBoundary.getIndex()
+      && panelAtLeftBoundary.getIndex() !== 0;
```
jongmoon: @WoodNeck Although it's not important and not critical. But it would be better to check `panelAtLeftBoundary.getIndex() !== 0` first and `panelAtLeftBoundary.getIndex() <= panelAtRightBoundary.getIndex()` later. 

Because
  1. if **0 check** is failed, it doesn't need to access getIndex of Right boundary.
  2. it would be more readable if more simple condition precedes.

src/components/Viewport.ts
```diff
@@ -953,26 +953,67 @@ export default class Viewport {
...
+
+    // Visible count of panel 0 on first screen
+    const panel0OnFirstscreen = Math.ceil((relativeHangerPosition + firstPanel.getSize() - relativeAnchorPosition) / sumOriginalPanelSize)
+      + Math.ceil((viewportSize - relativeHangerPosition + relativeAnchorPosition) / sumOriginalPanelSize)
+      - 1;
```
jongmoon: Is last `-1` calc for considering intersection of A and B?
  - A: `Math.ceil((relativeHangerPosition + firstPanel.getSize() - relativeAnchorPosition) / sumOriginalPanelSize)`
  - B: `Math.ceil((viewportSize - relativeHangerPosition + relativeAnchorPosition) / sumOriginalPanelSize)`

WoodNeck: That's right

src/components/Viewport.ts
WoodNeck: That looks good, I'll fix that.

Issue #239, opened 2019-06-10
[Flicking v2] Provide `rebuild` method for applying panel changes


jongmoon:
## Description
Flicking v2 cannot react the panel changes. So clone panels remain unchanged. 

By providing `rebuild` method, you can react panel changes by call this method.

## Steps to check or reproduce
  1. circular mode & React flicking
  2. update panels (using React way)
     - it will not update clones.

result: new panels and previous panels coexist.

PR #245, opened 2019-06-11
Add `rebuild()` method for panel update.

jongmoon:
## Issue
#239 

## Details
Add `rebuild()` method for panel update.


src/Flicking.js
```diff
@@ -384,6 +390,13 @@ export default class Flicking extends Mixin(Component).with(eventHandler) {
 		);
 	}
 
+	_clonePanel(v) {
+		const clone = v.cloneNode(true);
+
+		clone.classList.add(`${this.options.prefix}-clone`);
```
WoodNeck: I think `utils.classList` is more preferred for browser compatibility.

test/unit/methods.spec.js
```diff
@@ -109,7 +109,7 @@ describe("Methods call", function() {
 	describe("getPrevElement()", function() {
 		tutils.hooks.run();
 
-		// Given
+// Given
```
WoodNeck: blank spaces are removed

src/Flicking.js
```diff
@@ -1743,6 +1756,38 @@ export default class Flicking extends Mixin(Component).with(eventHandler) {
...
+
+		// panels' css values
+		this._initOriginalPanelStyle(panel.$list);
+
+		// Add clones
+		if (this._addClonePanels()) {
+			panel.$list = utils.toArray(this.$container.children);
+		}
+
+		this._setDefaultPanel(options.defaultIndex);
```
daybrush: @jongmoon 
Shouldn't `defaultIndex` be the current index when calling `rebuild`?

src/Flicking.js
```diff
@@ -1743,6 +1756,38 @@ export default class Flicking extends Mixin(Component).with(eventHandler) {
...
+	 *
+	 * @return {eg.Flicking} An instance of a module itself <ko>모듈 자신의 인스턴스</ko>
+	 */
+	rebuild() {
+		const panel = this._conf.panel;
+		const options = this.options;
+		const prefix = options.prefix;
+
+		// filter original panel (remove clones)
+		utils.toArray(this.$wrapper.querySelectorAll(`.${prefix}-clone`)).forEach(el => el.remove());
```
daybrush: @jongmoon 

IE does not support 'remove' methods.

I think you should use parentNode.removeChild.

src/Flicking.js
jongmoon: Thank you!

src/Flicking.js
jongmoon: @daybrush I think `rebuild` is used for working like initial time. Maintaining current index is policy issue. How about to add parameter?

```js
flicking.rebuild({defaultIndex: 1});
```

src/Flicking.js
jongmoon: Awesome!

src/Flicking.js
daybrush: @jongmoon 

When I heard, I knew that the contents of the panel were updated in the `circular`.

daybrush: @jongmoon

I think adding parameters is good.

PR #247, opened 2019-06-12
feat: add getNearestPanel()

### LINKED ISSUE #246: Add getNearestPanel
## Description
As currentPanel doesn't return cloned panels, nearest panel is needed if we have to use panel at hanger regardless of cloning.

WoodNeck:
## Issue
#246 

## Details
This resolves #246


src/Flicking.ts
```diff
@@ -293,6 +293,18 @@ class Flicking extends Component {
...
   }
 
+  /**
+   * Return nearest panel from hanger. `null` if no panel exists.
+   * @ko 행어로부터 가장 가까운 패널을 반환한다. 패널이 하나도 없을 경우 `null`을 반환한다.
+   * @return Nearest panel from hanger.<ko>행어로부터 가장 가까운 패널.</ko>
+   */
+  public getNearestPanel(): FlickingPanel | null {
+    const nearestPanel = this.viewport.getNearestPanel();
+    return nearestPanel
```
jongmoon: @WoodNeck How about ?
```js
return this.viewport.getNearestPanel() || null;
```

mixed:
@WoodNeck 
When will merge this pr?

WoodNeck:
@mixed 
I remember I've proposed this PR as it can be used in CFC structure, but it wasn't used eventually.
So, we've concluded that this PR is not so necessary at the moment.
It can be merged whenever it's needed, but it's not that time I think.
This is a quite short PR, so it can even be closed for sure.

PR #249, opened 2019-06-13
fix: resize() binding for window resize event

### LINKED ISSUE #248: Enabling autoResize option triggers error
## Description
Enabling Flicking's `autoResize` option triggers error.

image

Init Flicking with `autoResize` option, then change window size.

WoodNeck:
## Issue
#248

## Details
This fixes #248

### DIFF
--- a/src/Flicking.ts
+++ b/src/Flicking.ts
@@ -447,7 +447,7 @@ class Flicking extends Component {
    * @ko 패널들을 현재 상태에 맞춰 갱신한다.
    * @return {eg.Flicking} The instance itself.<ko>인스턴스 자기 자신.</ko>
    */
-  public resize(): this {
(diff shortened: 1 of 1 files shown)


src/Flicking.ts
```diff
@@ -447,7 +447,7 @@ class Flicking extends Component {
    * @ko 패널들을 현재 상태에 맞춰 갱신한다.
    * @return {eg.Flicking} The instance itself.<ko>인스턴스 자기 자신.</ko>
    */
-  public resize(): this {
+  public resize = (): this => {
```
jongmoon: @WoodNeck @daybrush Is it more preferable than `bind(this)`?

WoodNeck: I think it's more simple.

PR #251, opened 2019-06-17
fix: current panel can be cloned one

### LINKED ISSUE #250: Calling next() on first clone of panel 0 will move to previous panel in circular mode
## Description
If `next()` is called on first cloned panel 0, it will move to prev direction.

1. Init Flicking with `circular: true;`
2. Move to first cloned panel 0
3. Call `next()`

WoodNeck:
## Issue
#250 

## Details
This fixes #250

### DIFF
--- a/src/components/Viewport.ts
+++ b/src/components/Viewport.ts
@@ -727,7 +727,7 @@ export default class Viewport {
   }
 
   public setCurrentPanel(panel: Panel): void {
-    this.currentPanel = panel.getOriginalPanel();
+    this.currentPanel = panel;
   }
 
   public setLastIndex(index: number): void {


src/components/Viewport.ts
```diff
@@ -727,7 +727,7 @@ export default class Viewport {
   }
 
   public setCurrentPanel(panel: Panel): void {
-    this.currentPanel = panel.getOriginalPanel();
+    this.currentPanel = panel;
```
jongmoon: is #247 no more needed?

WoodNeck: #247 still can be useful while in animating, but it's not that important at the moment.

PR #255, opened 2019-06-19
fix: camera goes too far when bound is enabled

### LINKED ISSUE #219: Camera moves too far with bound option
## Description
May-22-2019 15-05-17
When moving to first panel at second panel(or last -1 to last), camera went too far like above gif.

That happens with `bound` option, and I think it might only occurs only when panel is small.

WoodNeck:
## Issue
#219 

## Details
This fixes #219

### DIFF
--- a/test/unit/initialize.spec.ts
+++ b/test/unit/initialize.spec.ts
@@ -796,6 +796,7 @@ describe("Initialization", () => {
       expect((plugin.destroy as sinon.SinonSpy).callCount).to.be.equals(1);
     });
   });
+
   describe("initialize component with decorator", () => {
     it("should check if the method of the class created with the decorator is properly entered.", () => {
       class TestFlicking {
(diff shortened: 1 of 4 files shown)


test/unit/initialize.spec.ts
```diff
@@ -822,4 +823,56 @@ describe("Initialization", () => {
...
+      // When
+      simulate(flickingInfo.element, {
+        deltaX: -40,
+        duration: 500,
+      });
+      await waitFor(1000);
+
+      // Then
+      const scrollArea = (flickingInfo.instance as any).viewport.getScrollArea();
+      expect(moves.every(move => move >= scrollArea.prev)).to.be.true;
```
jongmoon: @WoodNeck Is it more meaningful to check `scrollArea.next`?

test/unit/initialize.spec.ts
```diff
@@ -822,4 +823,56 @@ describe("Initialization", () => {
...
+      // When
+      simulate(flickingInfo.element, {
+        deltaX: 40,
+        duration: 200,
+      });
+      await waitFor(1000);
+
+      // Then
+      const scrollArea = (flickingInfo.instance as any).viewport.getScrollArea();
+      expect(moves.every(move => move <= scrollArea.next)).to.be.true;
```
jongmoon: @WoodNeck Is it more meaningful to check `scrollArea.prev`?

test/unit/initialize.spec.ts
WoodNeck: Because it's testing going out of bound when camera's at last panel to last - 1 panel.

test/unit/initialize.spec.ts
WoodNeck: Because it's testing going out of bound when camera's at last panel to last - 1 panel.

PR #277, opened 2019-08-08
feat: enhance resize performance with options

### LINKED ISSUE #276: Flicking's resize() is too slow when there're many panels
## Description
Flicking's `resize()` is very slow when there're many panels, as it recalculates all panel size individually.
I suggest adding 2 new options, `isEqualSize` and `isConstantSize`, which do
- isEqaulSize
  - type: boolean or list of the class name
  - This assumes all(boolean), or panels with the same class have same size for each other.
- isContantSize
  - type: boolean
  - This assumes all panels will have size after resize.

WoodNeck:
## Issue
#276 

## Details
- This adds 2 options, `isEqualSize` and `isConstantSize`
- This closes #276


src/Flicking.ts
```diff
@@ -471,18 +479,22 @@ class Flicking extends Component {
...
 
     const allPanels = viewport.panelManager.allPanels();
-    allPanels.forEach(panel => panel.unCacheBbox());
+    if (!options.isConstantSize) {
+      allPanels.forEach(panel => panel.unCacheBbox());
+    }
+
+    const shouldResetElements = options.renderOnlyVisible
+      && !options.isConstantSize
+      && options.isEqualSize !== true;
```
jongmoon: why use `options.isEqualSize !== true` instead of `!options.isEqualSize`?

src/components/Viewport.ts
```diff
@@ -1849,4 +1923,39 @@ export default class Viewport {
...
+        if (overlappedClass) {
+          panel.resize(panelBboxes[overlappedClass]);
+          panelBboxes[overlappedClass] = panel.getBbox();
+        } else {
+          panel.resize();
+        }
+      });
+      return;
+    }
+    panels.forEach(panel => {
```

Why is the panel resized once more?

src/Flicking.ts
jongmoon: @WoodNeck I could remind this

test/unit/options.spec.ts
```diff
@@ -993,4 +995,232 @@ describe("Initialization", () => {
...
+        panels[className]
+          ? panels[className].push(panel)
+          : panels[className] = [panel];
+        return panels;
+      }, {});
+
+      // Then
+      expect(flicking.getPanelCount()).equals(6);
+
+      // For panels with same class name, it shoould have same size
```
jongmoon: typo: shoould => should

src/utils.ts
```diff
@@ -109,6 +111,14 @@ export function addClass(element: HTMLElement, className: string): void {
   }
 }
 
+export function hasClass(element: HTMLElement, className: string): boolean {
+  if (element.classList) {
+    return element.classList.contains(className);
+  } else {
+    return (element.className.indexOf(className) < 0);
```
daybrush: @WoodNeck 

This method is very dangerous.

```html
<div class="class01 class02"></div>
```
```js
hasClass(el, "class");
```

Instead, use as below.
```js
element.className.split(" ").indexOf(className);
// or
element.className.match(new RegExp(`(\\s|^)${className}(\\s|$)`));
```

test/unit/options.spec.ts
WoodNeck: Thanks for checking!

src/components/Viewport.ts
WoodNeck: There're `return`s above.

src/components/Viewport.ts
```diff
@@ -1849,4 +1923,39 @@ export default class Viewport {
 
     return { removedPanels, addedPanels };
   }
+
+  private resizePanels(panels: Panel[]): void {
+    const options = this.options;
+    const panelBboxes = this.panelBboxes;
+
+    if (options.isEqualSize === true) {
```
daybrush: isEqualSize is string[] type automatically.

```ts
const isEqualSize = this.options.isEqualSize;

if (isEqualSize) {
    if (isEqualSize === true) {
    } else {
      // isEqualSize is string[] automatically
    }
}

Comment (WoodNeck) [reply src/components/Viewport.ts]: Looks like I just copy-pasted that, just deleting `as` worked for me.
```ts
    if (options.isEqualSize === true) {
      //
    } else if (options.isEqualSize) {
      // isEqualSize is string[]
    }
```

PR #280, opened 2019-08-19
feat: add fill() method for needPanel event

### LINKED ISSUE #279: Adding panel on needPanel event with circular option is difficult.
## Description
As `needPanel` event's `index` is based on a panel where it needs more panel before/after, if we have to insert panels when `circular` is enabled we have to check all 4 possibilities for correctness. The four possibilities emerge from combining whether it is on first / last panel with direction is prev or next.

Simply, by adding a function named `fill` at needPanel event can provide more convenience for the users.

WoodNeck:
## Issue
#279 

## Details
This resolves #279

### DIFF
--- a/src/components/Viewport.ts
+++ b/src/components/Viewport.ts
@@ -1749,17 +1749,29 @@ export default class Viewport {
           }
 
           const parsedElements = parseElement(element);
+          // Slice elements to fit size equal to empty spaces
           const elements = direction === DIRECTION.NEXT
             ? parsedElements.slice(0, indexRange.length)
-            : parsedElements.slice(parsedElements.length - indexRange.length);
-          const newPanels = direction === DIRECTION.NEXT
-            ? options.circular && index === panelManager.getLastIndex()
-              ? this.insert(0, elements)
-              : siblingPanel.insertAfter(elements)
-            : options.circular && index === 0
-              ? this.insert(indexRange.max - elements.length + 1, elements)
-              : siblingPanel.insertBefore(elements);
-          return newPanels;
+            : parsedElements.slice(-indexRange.length);
+
+          if (direction === DIRECTION.NEXT) {
+            if (options.circular && index === panelManager.getLastIndex()) {
+              // needPanel event is triggered on last index, insert at index 0
+              return this.insert(0, elements);
+            } else {
+              return siblingPanel.insertAfter(elements);
+            }
+          } else if (direction === DIRECTION.PREV) {
+            if (options.circular && index === 0) {
+              // needPanel event is triggered on first index(0), insert at the last index
+              return this.insert(indexRange.max - elements.length + 1, elements);
+            } else {
+              return siblingPanel.insertBefore(elements);
+            }
+          } else {
+            // direction is null when there're no panels exist
+            return this.insert(0, elements);
+          }
         },
       } as Partial<NeedPanelEvent>,
     );
--- a/src/components/Viewport.ts
+++ b/src/components/Viewport.ts
@@ -1487,7 +1487,7 @@ export default class Viewport {
     const viewportSize = state.size;
     const bounce = options.bounce;
 
-    let parsedBounce: number[] = bounce as [number, number];
+    let parsedBounce: number[];
     if (isArray(bounce)) {
       parsedBounce = (bounce as string[]).map(val => parseArithmeticExpression(val, viewportSize, DEFAULT_OPTIONS.bounce as number));
     } else {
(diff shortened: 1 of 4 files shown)


test/unit/events.spec.ts
```diff
@@ -827,6 +827,224 @@ describe("Events", () => {
...
+        flicking.on(EVENTS.NEED_PANEL, e => {
+          e.fill([
+            createHorizontalElement(20),
+            createHorizontalElement(20),
+            createHorizontalElement(20),
+            createHorizontalElement(20),
+            createHorizontalElement(20),
+          ]);
+        });
+        flicking.replace(0, createHorizontalElement(20));
```
jongmoon: What is an intend for replacing first item?

test/unit/events.spec.ts
```diff
@@ -827,6 +827,224 @@ describe("Events", () => {
...
+        flicking.on(EVENTS.NEED_PANEL, e => {
+          e.fill([
+            createHorizontalElement(20),
+            createHorizontalElement(20),
+            createHorizontalElement(20),
+            createHorizontalElement(20),
+            createHorizontalElement(20),
+          ]);
+        });
+        flicking.replace(5, createHorizontalElement(20));
```
jongmoon: Same above.

test/unit/events.spec.ts
WoodNeck: Flicking's panel info should be changed to trigger need panel event.
Therefore, I just replaced the first item, which isn't existing so basically it's the same as calling `append`.
I just used `replace` to match with the below test which must use `replace`.

jongmoon: @WoodNeck Thank you I could understand! How about commenting this ? I think it would be helpful~

WoodNeck: Sure, I'll do it right away.

test/unit/events.spec.ts
```diff
@@ -827,6 +827,224 @@ describe("Events", () => {
...
+
+        // Then
+        expect(needPanelEvents.length).equals(2);
```
jongmoon: Could you tell me why needPanelEvents occurs twice?

test/unit/events.spec.ts
```diff
@@ -827,6 +827,224 @@ describe("Events", () => {
...
+            createHorizontalElement(20),
+            createHorizontalElement(20),
+            createHorizontalElement(20),
+            createHorizontalElement(20),
+          ]);
+        });
+        flicking.replace(3, createHorizontalElement(100)); // should fill from 3 to 5
+
+        // Then
+        expect(needPanelEvents.length).equals(2);
```
jongmoon: Same above.

test/unit/events.spec.ts
WoodNeck: This test is testing that flicking can fill panels in empty positions, which is, 1 to 4.
So as the event handler for `needPanel` is already set in `createInfiniteFlickingWithOption` to store `needPanel `events in `needPanelEvents` variable, when I call `flicking.replace(0)` in line **974** triggers `needPanel` event, so that's the one. The other is when I call `replace(5)`, so that's two of them.

test/unit/events.spec.ts
WoodNeck: Same as above, `needPanel` is triggered both in `flicking.replace(5)` and `flicking.replace(3)` respectively in the line 1015 and 1032.

test/unit/events.spec.ts
jongmoon: @WoodNeck  I got it! I was mistaken that handler for `EVENTS.NEED_PANEL` was registered after 974 line. But I became to know `needPanelEvents` are handled in other part.

src/components/Viewport.ts
```diff
@@ -1741,6 +1742,25 @@ export default class Viewport {
         panel: siblingPanel,
         direction,
         range: indexRange,
+        fill: (element: ElementLike | ElementLike[]) => {
+          const panelManager = this.panelManager;
+          if (!siblingPanel) {
+            return this.insert(panelManager.getRange().max + 1, element);
+          }
+
+          const parsedElements = parseElement(element);
```
It's hard to read because there are too many `?`. I think it's better to change it to `if` phrase.

src/components/Viewport.ts
```diff
@@ -1741,6 +1742,25 @@ export default class Viewport {
...
+          const elements = direction === DIRECTION.NEXT
+            ? parsedElements.slice(0, indexRange.length)
+            : parsedElements.slice(parsedElements.length - indexRange.length);
```
daybrush:
```js
parsedElements.slice(-indexRange.length);
```

WoodNeck: That's cool. I should use that.

src/components/Viewport.ts
WoodNeck: I agree that can be burdensome to read. I'll fix that.

Issue #273, opened 2019-08-05
Apply renderOnlyVisible option to frameworks


WoodNeck:
## Description
Apply "renderOnlyVisible" made in #272 to each framework.

PR #281, opened 2019-08-22
fix: match framework rendering with native rendering

WoodNeck:
## Issue
#273

## Details
This PR is made to match the rendering method same with native's.
That is, when a new panel threw in, the framework should render new panels with visible ones to calculate size.
This also implements `isEqualSize` and `isConstantSize` options for framework Flickings.
The real implementation is only available for only Vue-Flicking yet.


WoodNeck:
Close this PR due to it's not updating.
Tried https://github.com/isaacs/github/issues/750#issuecomment-275214525, but didn't work.
image
^ This is what I've got.

PR #282, opened 2019-08-22
fix: match framework rendering with native rendering

WoodNeck:
## Issue
#273

## Details
This PR is made to match the rendering method same with native's.
That is, when a new panel threw in, the framework should render new panels with visible ones to calculate size.
This also implements `isEqualSize` and `isConstantSize` options for framework Flickings.
The real implementation is only available for only Vue-Flicking yet.

### DIFF
--- a/packages/vue-flicking/src/Flicking.ts
+++ b/packages/vue-flicking/src/Flicking.ts
@@ -3,7 +3,7 @@
  * egjs projects are licensed under the MIT license
  */
 
-import NativeFlicking, { Plugin, FlickingOptions, DestroyOption, withFlickingMethods, DEFAULT_OPTIONS, FlickingPanel } from "../../../src/index";
+import NativeFlicking, { Plugin, FlickingOptions, DestroyOption, withFlickingMethods, DEFAULT_OPTIONS, FlickingPanel } from "@egjs/flicking";
 import ChildrenDiffer from "@egjs/vue-children-differ";
 import ListDiffer, { DiffResult } from "@egjs/list-differ";
 import { Component, Vue, Prop } from "vue-property-decorator";
@@ -27,15 +27,11 @@ export default class Flicking extends Vue {
   private $_pluginsDiffer!: ListDiffer<Plugin>;
   private $_cloneCount!: number;
   private $_slotDiffer!: ListDiffer<VNode>;
-  private $_slotDiffResult: DiffResult<VNode> | undefined;
-  private $_visibleIndex!: { min: number; max: number };
 
   public mounted() {
     this.$_pluginsDiffer = new ListDiffer<Plugin>();
     this.$_cloneCount = 0;
-    this.$_visibleIndex = { min: NaN, max: NaN };
 
-    this.options.renderOnlyVisible = true;
     const options = {...this.options, ...{ renderExternal: true }};
     this.$_nativeFlicking = new NativeFlicking(this.$el as HTMLElement, options);
     this.$_slotDiffer = new ListDiffer<VNode>(this.$slots.default, (vnode, idx) => vnode.key!);
@@ -45,11 +41,6 @@ export default class Flicking extends Vue {
 
     if (this.options.renderOnlyVisible) {
       // Should update once to update visibles
-      const visibleIndex = this.$_nativeFlicking.getVisibleIndex();
-      this.$_visibleIndex = {
-        min: visibleIndex.min,
-        max: visibleIndex.max,
-      };
       this.$forceUpdate();
     }
   }
@@ -100,7 +91,6 @@ export default class Flicking extends Vue {
 
     this.$_nativeFlicking.sync(diffResult);
     this.$_cloneCount = newCloneCount;
-    this.$_slotDiffResult = undefined;
     this.$nextTick(() => {
       this.$_checkUpdate();
     });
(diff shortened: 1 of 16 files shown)


packages/vue-flicking/src/Flicking.ts
```diff
@@ -28,29 +27,20 @@ export default class Flicking extends Vue {
...
   public mounted() {
     this.$_pluginsDiffer = new ListDiffer<Plugin>();
     this.$_cloneCount = 0;
-    this.$_visibleIndex = { min: NaN, max: NaN };
 
-    this.options.renderOnlyVisible = true;
     const options = {...this.options, ...{ renderExternal: true }};
     this.$_nativeFlicking = new NativeFlicking(this.$el as HTMLElement, options);
-    this.$_slotDiffer = new ListDiffer<VNode>(this.$slots.default, (vnode, idx) => vnode.key || idx);
+    this.$_slotDiffer = new ListDiffer<VNode>(this.$slots.default, (vnode, idx) => vnode.key!);
```
jongmoon: Why index is removed?

packages/vue-flicking/src/Flicking.ts
```diff
@@ -157,74 +143,37 @@ export default class Flicking extends Vue {
 
   private $_getPanels(h: CreateElement) {
     const flicking = this.$_nativeFlicking;
```
jongmoon: @WoodNeck More readable than before 👍

packages/vue-flicking/src/Flicking.ts
WoodNeck: Because vnode's key is guaranteed.

```js
  public beforeUpdate() {
    if (!this.$slots.default) {
      return;
    }

    this.$slots.default = this.$slots.default.map((node, index) => {
      if (node.key == null) {
        node.key = `${node.tag}-${index}`;
      }
      return node;
    });
  }
```

src/Flicking.ts
```diff
@@ -582,24 +576,43 @@ class Flicking extends Component {
...
-   * @ko 현재 렌더링 중인 패널들에 대해 사용자의 전체 아이템들과 매핑을 시킨다.
-   * @param - User's all items. <ko>사용자의 전체 아이템들.</ko>
-   * @return Array of Item of the user mapped to the currently rendering panels. <ko>현재 렌더링 중인 패널과 매핑되는 사용자의 아이템 배열.</ko>
+   * @ko 렌더링이 필요한 인덱스들을 반환한다. `renderOnlyVisible` 옵션과 함께 사용해야 한다.
+   * @param - Info object of how panel infos are changed.<ko>패널 정보들의 변경 정보를 담는 오브젝트.</ko>
+   * @return Array of indexes to render.<ko>렌더링할 인덱스의 배열</ko>
    */
-  public mapRenderingPanels<T = any>(allItems: T[]): T[] {
-    const { renderOnlyVisible } = this.options;
+  public getRenderingIndexes(diffResult: DiffResult<any>): number[] {
```
jongmoon: I also need index more than panels.

src/Flicking.ts
```diff
@@ -677,8 +697,7 @@ class Flicking extends Component {
       maintained.forEach(([, next]) => { viewport.updateCheckedIndexes({ min: next, max: next }); });
     }
     panelManager.replacePanels(newPanels, newClones);
-
-    !isCallSync && this.viewport.beforeSync();
```
jongmoon: How `vieweport.beforeSync` can be removed? What is the replacement of `viewerport.beforeSync`?

src/components/Viewport.ts
```diff
@@ -995,19 +992,25 @@ export default class Viewport {
       panels.forEach(panel => {
         const overlappedClass = panel.getOverlappedClass(equalSizeClasses);
         if (overlappedClass && !cached[overlappedClass]) {
-          fragment.appendChild(panel.getElement());
+          if (!options.renderExternal) {
```
jongmoon: Omission of checking `renderExternal` was a bug?

src/Flicking.ts
```diff
@@ -677,8 +697,7 @@ class Flicking extends Component {
...
+    this.flag = true;
```
jongmoon: `this.flag` should have meaningful name.

WoodNeck: Oops I forgot to change that one.

src/components/Viewport.ts
WoodNeck: Yes, that made `flicking.resize()` not to work.

src/components/Panel.ts
```diff
@@ -299,26 +299,32 @@ class Panel implements FlickingPanel {
...
       state.cachedBbox = {
         x: 0,
         y: 0,
         width: 0,
         height: 0,
       };
     } else if (!state.cachedBbox) {
-      if (!this.element.parentNode && !viewport.options.renderExternal) {
-        viewport.getCameraElement().appendChild(this.element);
+      const wasVisible = Boolean(element.parentNode);
```
jongmoon: @WoodNeck It's very trivial question. I think `isVisible` is more proper because it has parentNode currently. Why do you use `was`?

WoodNeck: Well, as it inserts a panel element when it wasn't visible, and as I should find panel elements that I've inserted at that sequence to remove it (that is, "wasn't visible"), I thought `wasVisible` is more proper.

src/Flicking.ts
WoodNeck: That causes the case I've told you that call count of render() increases.
Because `viewport.beforeSync` calls updateVisiblePanels, and as it can trigger the `visibleChange` event, it can increase rendering cost about twice as much.
I've handled that by using `this.flag` (I'll change its name soon).

PR #285, opened 2019-08-23
fix: rendering method for React

WoodNeck:
## Issue
#273

## Details
As same with #282, this matches rendering method on React the same with native Flicking.

### DIFF
--- a/src/components/Viewport.ts
+++ b/src/components/Viewport.ts
@@ -188,7 +188,7 @@ export default class Viewport {
     const scrollArea = state.scrollArea;
 
     // Update position & nearestPanel
-    if (!isBetween(pos, scrollArea.prev, scrollArea.next)) {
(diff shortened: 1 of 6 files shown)


src/components/Viewport.ts
```diff
@@ -1781,11 +1781,8 @@ export default class Viewport {
       const { addedPanels, removedPanels } = this.checkVisiblePanelChange(prevVisiblePanels, newVisiblePanels);
 
       if (newVisiblePanels.length > 0) {
-        const firstPanelPos = this.panelManager.firstPanel()!.getPosition();
         const firstVisiblePanelPos = newVisiblePanels[0].getPosition();
-        if (firstVisiblePanelPos >= firstPanelPos) {
```
jongmoon: @WoodNeck Is always true?

WoodNeck: Yes, as it sets firstVisiblePanelPos on position 0

Issue #286, opened 2019-08-27
flicking.rebuild ({defaultIndex: 0}) is not working.


daybrush:
## Description

Other indexes work fine, but flicking.rebuild ({defaultIndex: 0}) doesn't work.

PR #287, opened 2019-08-27
fix: fix defaultIndex is zero

daybrush:
## Issue

#286


src/Flicking.js
```diff
@@ -469,7 +469,7 @@ export default class Flicking extends Mixin(Component).with(eventHandler) {
 				currNo: index
 			});
 			// if defaultIndex option is given, then move to that index panel
-		} else if (index > 0 && index <= lastIndex) {
+		} else if (((!isRebuild && index > 0) || (isRebuild && index >= 0)) && index <= lastIndex) {
```
jongmoon: @daybrush 2 Question

1. This code is not needed to be applied on `circular`mode?
2. `isRebuild` is needed? How about following?
  - `if (index !== prevIndex && index >= 0 && index <= lastIndex)`

daybrush: @jongmoon

2.
There will be many different cases where the user rebuilds.
Even if index does not change, the previewPadding or width changes.

PR #290, opened 2019-09-02
feat: add useOffset

### LINKED ISSUE #289: Panel size is miscalculated when CSS transform is applied
## Description
As Flicking internally uses `getBoundingClientRect()`, panel size can be miscalculated when CSS `transform` is applied to panel or wrapper.
`getBoundingClientRect` calculates element size after transform is applied, while `offsetWidth` returns original size.
So basically if we use `offsetWidth` instead of `getBoundingClientRect` we can get original panel size, regardless of transform. However, offsetWidth returns value in integer so there can be 1px gap between panels if the panel's size includes decimal fraction.

Therefore we're going to make a new option called `useOffset` to use `offsetWidth` and `offsetHeight` instead of `getBoundingClientRect`, so you can choose the way to calculate panel size.

WoodNeck:
## Issue
#289 

## Details
This resolves #289


src/components/Viewport.ts
```diff
@@ -608,14 +608,26 @@ export default class Viewport {
...
+
     if (!state.cachedBbox) {
-      const bbox = this.viewportElement.getBoundingClientRect();
-      state.cachedBbox = {
-        x: bbox.left,
-        y: bbox.top,
-        width: bbox.width,
-        height: bbox.height,
-      };
+      if (options.useOffset) {
```
jongmoon: @WoodNeck This code is almost same with Panels' code. Is there any solution to use it commonly.

WoodNeck: I'll refactor that, thanks.

src/components/Viewport.ts
```diff
@@ -608,14 +608,11 @@ export default class Viewport {
...
+
     if (!state.cachedBbox) {
-      const bbox = this.viewportElement.getBoundingClientRect();
-      state.cachedBbox = {
-        x: bbox.left,
-        y: bbox.top,
-        width: bbox.width,
-        height: bbox.height,
-      };
+      state.cachedBbox = getBbox(viewportElement, options.useOffset, true);
```
I think x and y will look very different in the `calcNewVisiblePanelIndex` method. Is that okay?

src/Flicking.ts
```diff
@@ -93,6 +93,7 @@ class Flicking extends Component {
...
    * @param {number} [options.gap=0] Space between each panels. Should be given in number.(px).<ko>패널간에 부여할 간격의 크기를 나타내는 숫자.(px)</ko>
    * @param {eg.Flicking.MoveTypeOption} [options.moveType="snap"] Movement style by user input.(ex: snap, freeScroll)<ko>사용자 입력에 의한 이동 방식.(ex: snap, freeScroll)</ko>
+   * @param {boolean} [options.useOffset=false] Whether to use `offsetWidth`/`offsetHeight` instead of `getBoundingClientRect` for panel/viewport size calculation.<br/>You can use this option to calculate the original panel size when CSS transform is applied to viewport or panel.<br/>⚠️ If panel size is not fixed integer value, there can be a 1px gap between panels.<ko>패널과 뷰포트의 크기를 계산할 때 `offsetWidth`/`offsetHeight`를 `getBoundingClientRect` 대신 사용할지 여부.<br/>패널이나 뷰포트에 CSS transform이 설정되어 있을 때 원래 패널 크기를 계산하려면 활성화할 수 있다.<br/>⚠️ 패널의 크기가 정수로 고정되어있지 않다면 패널 사이에 1px의 공간이 생길 수 있다.</ko>
```
jongmoon: The sentence `원래 패널 크기를 계산하려면 활성화 할 수 있다.`  is a bit awkward. How about `원래 패널 크기를 계산하려면 옵션을 활성화 한다.`

src/utils.ts
```diff
@@ -301,3 +301,31 @@ export function withFlickingMethods(prototype: any, flickingName: string) {
     };
   });
 }
+
+export function getBbox(element: HTMLElement, useOffset: boolean, includeXY: boolean) {
+  let bbox: BoundingBox;
+  if (useOffset) {
+    bbox = includeXY
```
daybrush: @WoodNeck 

The code is long. How about this?
```js
const x = includeXY ? ...
const y = includeXY ? ...
```

src/components/Viewport.ts
WoodNeck: That looks not good, I think I should fix that.

test/unit/options.spec.ts
```diff
@@ -1204,4 +1204,74 @@ describe("Initialization", () => {
...
+      // When
+      flicking.resize();
+
+      // Then
+      flicking.getAllPanels().forEach(panel => {
+        const panelSize = panel.getSize();
+        expect(panelSize).not.equals(parseInt(panelSize.toString(), 10));
+      });
+    });
+  });
```
jongmoon: @WoodNeck I think it would be better to add test case that checks `return original panel size only if `useOffset=true` although transform is applied.`

src/utils.ts
WoodNeck: Sometimes, I really am obsessed with such trivial things I think.
Actually I just wrote code like that to remove one single branch, which is `includeXY` in this case.
But why not, I think 2 `includeXY` is not bad.

src/Flicking.ts
WoodNeck: Sure, I'll fix that.

test/unit/options.spec.ts
WoodNeck: Okay, I'll add that.

PR #291, opened 2019-09-05
Apply renderOnlyVisible option to frameworks 

jongmoon:
## Issue
#273 

## Details
  - Apply `renderOnlyVisible` option.
  - (WorkingInProcess): supplement documentation & demo.

### DIFF
--- a/packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
+++ b/packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
@@ -14,6 +14,7 @@ export interface RenderPanelChangeEvent {
   visibles: {
     index: number;
     key: number; /* Unique value */
+    data: any; /* User panel data */
   }[];
 }
 
@@ -217,8 +218,12 @@ export class NgxFlickingComponent implements OnInit, AfterViewInit, OnDestroy, O
     this.prevVisibles = visibles;
 
     const l = this.panels.length;
-    const renderChangeEvent = {
-      visibles: visibles.map(i => ({key: i, index: i % l}))
+    const renderChangeEvent: RenderPanelChangeEvent = {
+      visibles: visibles.map(i => ({
+        key: i,
+        index: i % l,
+        data: this.panels[i % l]
+      }))
     };
 
     if (Promise) {
--- a/packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
+++ b/packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
@@ -1,13 +1,17 @@
+// import { VisibleChangeEvent } from './../../../../../../declaration/types.d';
+// import { Flicking } from '@egjs/flicking';
 /**
  * Copyright (c) 2015 NAVER Corp.
  * egjs projects are licensed under the MIT license
  */
 
 // tslint:disable-next-line: max-line-length
-import NativeFlicking, { Plugin, FlickingOptions, withFlickingMethods, DEFAULT_OPTIONS, FlickingEvent, NeedPanelEvent, SelectEvent, ChangeEvent } from '@egjs/flicking';
+import NativeFlicking, { Plugin, FlickingOptions, withFlickingMethods, DEFAULT_OPTIONS, FlickingEvent, NeedPanelEvent, SelectEvent, ChangeEvent, VisibleChangeEvent } from '../../../../../../src/index';
+import { counter } from '../../../../../../src/utils';
 // tslint:disable-next-line: max-line-length
-import { Component, OnInit, Input, AfterViewInit, ElementRef, OnChanges, Output, EventEmitter, OnDestroy, ContentChild, TemplateRef } from '@angular/core';
-import ListDiffer, { DiffResult } from '@egjs/list-differ';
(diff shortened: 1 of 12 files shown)


packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
```diff
@@ -4,9 +4,10 @@
  */
 
 // tslint:disable-next-line: max-line-length
-import NativeFlicking, { Plugin, FlickingOptions, withFlickingMethods, DEFAULT_OPTIONS, FlickingEvent, NeedPanelEvent, SelectEvent, ChangeEvent } from '@egjs/flicking';
+import NativeFlicking, { Plugin, FlickingOptions, withFlickingMethods, DEFAULT_OPTIONS, FlickingEvent, NeedPanelEvent, SelectEvent, ChangeEvent, VisibleChangeEvent } from '@egjs/flicking';
+import { counter } from '../../../../../../src/utils';
```
WoodNeck: If we're not targeting IE8, I think it's better to use another implementation of counter(or `range`, like python).
As native Flicking's counter is quite rough.

WoodNeck: Like,
```js
const counter = num => Array(num).map((_, i) => i));
// or 
const counter = num => [...Array(num).keys()];
```

packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
```diff
@@ -27,9 +28,10 @@ import ListDiffer, { DiffResult } from '@egjs/list-differ';
     ':host {display: block}', ':host > div {width: 100%; height: 100%}'
   ]
 })
-export class NgxFlickingComponent implements OnInit, AfterViewInit, OnDestroy, OnChanges {
+export class NgxFlickingComponent implements OnInit, AfterViewInit, OnDestroy, OnChanges, AfterViewChecked, DoCheck {
   @Input() options: Partial<FlickingOptions> = {};
   @Input() plugins: Plugin[] = [];
+  @Input() panels: [];
```
WoodNeck: I think this code regards the type of `panels` as an empty array.
If you intended to set `panels` can be anything, I think `any[]` is a more proper way to do that.

src/components/Viewport.ts
```diff
@@ -1818,7 +1818,7 @@ export default class Viewport {
       }
 
       newVisiblePanels.forEach(panel => {
-        panel.setPositionCSS(state.positionOffset);
+          panel.setPositionCSS(state.positionOffset);
```
WoodNeck: Indent is doubled here

packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
daybrush: Array.keys also cannot support IE.

packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
```diff
@@ -133,8 +183,38 @@ export class NgxFlickingComponent implements OnInit, AfterViewInit, OnDestroy, O
...
+    }
+
+    this.prevVisibles = visibles;
+
+    if (this.criticalSection) {
+      // Use microtask to run as soon as synchronous job is done.
+      Promise.resolve()
+        .then(() => this.renderPanelChange.emit(visibles));
+    } else {
+      this.renderPanelChange.emit(visibles);
```
daybrush: @jongmoon 
What about sending events as objects rather than as arrays?

```js
emit({
    visibles,
});

File: packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts (hunk 1 above)
Comment (WoodNeck) [reply packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts]: I mean, **If we're not targeting IE8** in Angular.

Comment (jongmoon) [reply packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts]: I've just found that angular support from IE9 ~ 😢 
https://angular.io/guide/browser-support
image

File: packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts (hunk 2 above)
Comment (jongmoon) [reply packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts]: @WoodNeck Yes, panels can be anything as you said. I'll fit it. Thank you!

File: src/components/Viewport.ts (hunk 3 above)
Comment (jongmoon) [reply src/components/Viewport.ts]: @WoodNeck My mistake!

File: packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts (hunk 1 above)
Comment (WoodNeck) [reply packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts]: I know it's trivial, but the upper one can be used instead.
Because native Flicking's counter is like this.
```ts
// return [0, 1, ...., max - 1]
export function counter(max: number): number[] {
  const counterArray: number[] = [];
  for (let i = 0; i < max; i += 1) {
    counterArray[i] = i;
  }
  return counterArray;
}
```
It's made in the beginning of 3.0, when I had no clue whether to support IE8 or not.

packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
jongmoon: @daybrush I try to change it. check it please after I pull request.

PR #288, opened 2019-08-29
docs: update 3.4.0 documents

WoodNeck:

PR #293, opened 2019-09-05
test: apply fake timer

### LINKED ISSUE #292: Test is too slow & is broken sometimes
## Description
Our test cases are very slow(approx, 150sec), and can be broken anytime with the case of floating-point issue(https://github.com/naver/egjs-axes/pull/141).
If we use fake timers instead, we can achieve both test time reduction & test accuracy as we can control intervals between frames.

WoodNeck:
## Issue
#292

## Details
This resolves #292

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,7 +5,7 @@ root = true
 insert_final_newline = true
 trim_trailing_whitespace = true
 
-[*.ts, *.vue]
+[*.ts, *.js, *.vue]
 charset = utf-8
 end_of_line = lf
 indent_style = space
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -11,6 +11,7 @@ module.exports = function(config) {
     },
     files: [
       "./node_modules/hammer-simulator/index.js",
+      "./test/setup.js",
       "./test/hammer-simulator.run.js",
       "./src/**/*.ts",
       "./test/**/*.ts",
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,59 +1,59 @@
 module.exports = function(config) {
-	const karmaConfig = {
-		frameworks: ["mocha", "chai", "sinon", "karma-typescript", "viewport"],
-		mime: {
-      		'text/x-typescript': ['ts','tsx']
-		},
-		client: {
-			mocha: {
-				opts: "./mocha.opts",
-			},
-		},
-		files: [
-			"./node_modules/hammer-simulator/index.js",
-			"./test/hammer-simulator.run.js",
-			"./src/**/*.ts",
-			"./test/**/*.ts",
-			"./test/unit/assets/*.css",
-		],
-		preprocessors: {
-			"src/**/*.ts": ["karma-typescript"],
-			"test/**/*.ts": ["karma-typescript"],
-		},
-		karmaTypescriptConfig: {
-			tsconfig: "./tsconfig.test.json",
-			reports: {
-        		html: {
-					"directory": "coverage",
-					"subdirectory": "./"
-				},
-				lcovonly: {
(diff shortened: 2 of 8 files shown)


karma.conf.js
```diff
@@ -1,59 +1,60 @@
...
-		browsers: [],
-		customLaunchers: {
-			CustomChromeHeadless: {
-				base: "ChromeHeadless",
-				flags: ["--window-size=400,300", "--no-sandbox", "--disable-setuid-sandbox"],
-			},
-		},
-		reporters: ["mocha"],
-	};
+  const karmaConfig = {
```
WoodNeck: Nothing changed here but changed indent to space-2

test/unit/assets/utils.ts
```diff
@@ -65,8 +64,12 @@ export function cleanup() {
   });
 }
 
+export function tick(time) {
+  (window as any).timer.tick(time);
+}
+
 declare var Simulator: any;
-export function simulate(el: HTMLElement, option?: object): Promise<void> {
+export function simulate(el: HTMLElement, option?: object, time: number = 15000): Promise<void> {
```
jongmoon: Isn't it working synchronously?

test/unit/events.spec.ts
```diff
@@ -504,18 +467,17 @@ describe("Events", () => {
...
 
     it("will trigger events in correct order even if interrupted", async () => {
-      await simulate(flickingInfo.element, { deltaX: -100, duration: 50 });
-      await waitFor(25);
-      await simulate(flickingInfo.element, { deltaX: -100, duration: 100 });
-      await waitFor(1000);
+      // Don't let first simulate finished, as this is testing interruptablility
+      // Setting time same to duration(400) makes only panning behavior finished, but not animation
+      simulate(flickingInfo.element, { deltaX: -500, duration: 400 }, 400);
+      await simulate(flickingInfo.element, { deltaX: -500, duration: 100 });
```
jongmoon: is `await` needed?

.editorconfig
```diff
@@ -5,7 +5,7 @@ root = true
 insert_final_newline = true
 trim_trailing_whitespace = true
 
-[*.ts, *.vue]
+[*.ts, *.js, *.vue]
```
daybrush: @WoodNeck 

Does it work?

daybrush: @WoodNeck 
Does it work?

WoodNeck: It doesn't. I've fixed it in a new commit.

test/unit/assets/utils.ts
WoodNeck: Actually it is.
But I just let this function asynchronous as hammer-simulator calls `done()` after panning is finished.
I think it can be used more practically when we return Promise in this function. Because we can assume panning is finished when the promise is resolved.

test/unit/events.spec.ts
WoodNeck: All `await `is not necessary now.
It isn't important in all of the current tests, but I just left that because in this test we're testing whether interrupting while animating is permitted.
My purpose was to differentiate them. The first one should not be finished, while the second one must be finished.
It has no meaning in the result as you said.

test/unit/assets/utils.ts
jongmoon: Ok, I got it.

test/unit/events.spec.ts
jongmoon: @WoodNeck I understand your intent. And I think it's good intent. But what I am concerned is whether we can keep this intention in the future.

.editorconfig
WoodNeck: commit is added in #288, my mistake.

PR #306, opened 2019-10-01
feat: collect information with GA

### LINKED ISSUE #301: Omitted framework information for GA (Vue, Angular)
## Description
Framework information is omitted on Angular, Vue.

Reference React Implementation
https://github.com/naver/egjs-flicking/blob/1fd1ccb584a29920fe7089828197f3e02bfceeac/packages/react-flicking/src/react-flicking/Flicking.tsx#L76

TroyTae:
## Issue

fixes #301

## Details

I defined third parameter to collect information with GA.
This code is fine for framework users, but pure js user can pass useless things through third parameter. (This issue is already existed)
So before merge this PR, we need to think about how prevention that.

### DIFF
--- a/packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
+++ b/packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
@@ -3,10 +3,34 @@
  * egjs projects are licensed under the MIT license
  */
 
-// tslint:disable-next-line: max-line-length
-import NativeFlicking, { Plugin, FlickingOptions, withFlickingMethods, DEFAULT_OPTIONS, FlickingEvent, NeedPanelEvent, SelectEvent, ChangeEvent, VisibleChangeEvent } from '@egjs/flicking';
-// tslint:disable-next-line: max-line-length
-import { Component, OnInit, Input, AfterViewInit, ElementRef, OnChanges, Output, EventEmitter, OnDestroy, ContentChild, TemplateRef, SimpleChanges, AfterViewChecked, DoCheck } from '@angular/core';
+import NativeFlicking, {
+  Plugin,
+  FlickingOptions,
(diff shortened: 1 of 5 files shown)


TroyTae:
Hello, @WoodNeck 

I want to achieve two goals:

1. Correct type
As you said, I don't like to use `any` type.
And I think framework/frameworkVersion isn't related with `FlickingOption`.
Of course, using third parameter just for GA looks bad!
How about leaving the third parameter as a universal variable for internal use?

2. Do not pass useless data to GA
Even if we use `options` instead of third parameter, non-typescript user can pass useless properties.
So I suggest you, how about using spread operator?
```
constructor(
  element: string | HTMLElement,
  options: Partial<FlickingOptions> = {},
  {framework, frameworkVersion}: Partial<CollectInformation> = {},
) {
  sendEvent(
    "usage",
    "options",
    {
      ...options,
      framework,
      frameworkVersion,
    },
  );
}
```

WoodNeck:
@TroyTae 
I get your point, and that's quite reasonable I think.
But the thing is, we should change the interface very carefully as there's no turning back when there's a need for a non-internal parameter, especially as it's a constructor interface change.
Also, as we're trying to hide framework-related options & methods as possible like you can see the document of `sync` method, adding a parameter which only used internally in the constructor seems quite unreasonable for me.
While you may hate some useless properties can be in the options object when GA send occurs, actually we found out that it's pretty acceptable. Even deprecated options of version 2.x are still sent, and that's still useful to us.
If handling useless properties become burdensome, we can rather change the part where we send options to check keys are in `FlickingOptions` type.

TroyTae:
@WoodNeck 
It makes sense!
I think about it again, using constructor parameter for internal variables seems bad.
I'll reflect it :)

WoodNeck:
I'll merge this PR when @daybrush finishes the review.

PR #312, opened 2019-10-31
fix: fix script error for OnPush mode

daybrush:
## Issue

#311

### DIFF
--- a/packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
+++ b/packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
@@ -30,6 +30,8 @@ import {
   AfterViewChecked,
   DoCheck,
   VERSION,
+  ChangeDetectionStrategy,
+  ChangeDetectorRef,
 } from '@angular/core';
 import ListDiffer, { DiffResult } from '@egjs/list-differ';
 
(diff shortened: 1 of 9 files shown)


packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
```diff
@@ -201,7 +201,7 @@ export class NgxFlickingComponent implements OnInit, AfterViewInit, OnDestroy, O
     }
 
     this.internalCloneCount = newCC;
-    this.cdr.detectChanges();
+    this.cdr.markForCheck();
```
WoodNeck: What's the reason using `markForCheck` instead of `detectChanges`?

daybrush: @WoodNeck 

markForCheck(): Marks all ChangeDetectionStrategy ancestors as to be checked.
detectChanges(): Checks the change detector and its children.

https://medium.com/@pablo.platt/can-you-please-explain-when-to-use-markforcheck-and-when-detectchanges-52e281783c6

WoodNeck: I wanted you to explain the reason why you've chosen `markForCheck`, not definition.
Looks like this link explains good about it.
https://stackoverflow.com/questions/41364386/whats-the-difference-between-markforcheck-and-detectchanges/45396740#41364469

PR #322, opened 2019-11-27
fix: scrollArea calculation when sum of the panel sizes is smaller than the viewport size

### LINKED ISSUE #321: Flicking can go out of bound with bound option when there're few panels.
## Description
When there're few panels so that sum of panel sizes can't reach viewport size, flicking will act like bound options is false.
So, when there's a change in viewport area size like a rotating the mobile device, it can act like a bound option is turning on and off.
By making Flicking's scroll area to 0 in such condition can resolve this issue.

WoodNeck:
## Issue
#321

## Details
This fixes #321

### DIFF
--- a/src/components/Viewport.ts
+++ b/src/components/Viewport.ts
@@ -1434,7 +1434,7 @@ export default class Viewport {
         next: 0,
       };
     } else if (this.canSetBoundMode()) {
-      const sumOriginalPanelSize = lastPanel.getPosition() + lastPanel.getSize() - firstPanel.getPosition() + options.gap;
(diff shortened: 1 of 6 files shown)


src/components/Viewport.ts
```diff
@@ -1443,10 +1434,27 @@ export default class Viewport {
         next: 0,
       };
     } else if (this.canSetBoundMode()) {
-      state.scrollArea = {
-        prev: firstPanel.getPosition(),
-        next: lastPanel.getPosition() + lastPanel.getSize() - state.size,
-      };
+      const sumOriginalPanelSize = lastPanel.getPosition() + lastPanel.getSize() - firstPanel.getPosition() + options.gap;
```

Why add options.gap?

WoodNeck: That's my mistake, thanks for checking it.

PR #328, opened 2019-11-28
fix: text node can be used as panel

### LINKED ISSUE #327: Vue - element.getAttribute is not a function
## Description
image
```
[Vue warn]: Error in directive children-differ componentUpdated hook: "TypeError: element.getAttribute is not a function"

found in

---> <Flicking>
       <KeywordScore> at src/components/KeywordScore.vue
         <App> at src/App.vue
           <Root>

vue.esm.js?efeb:1897 TypeError: element.getAttribute is not a function
    at Panel.__proto.setElement (flicking.esm.js?1454:902)
    at eval (flicking.esm.js?1454:4917)
    at Array.forEach (<anonymous>)
    at Flicking.__proto.sync (flicking.esm.js?1454:4914)
    at VueComponent.__proto.onUpdate (flicking.esm.js?bcd9:184)
    at componentUpdated (children-differ.esm.js?2cdd:25)
    at callHook$1 (vue.esm.js?efeb:6685)
    at eval (vue.esm.js?efeb:6637)
    at wrappedHook (vue.esm.js?efeb:2244)
    at invokeWithErrorHandling (vue.esm.js?efeb:1863)
```

This can happen when the text node is used for the Flicking panel.

WoodNeck:
## Issue
#327

## Details
This fixes #327

### DIFF
--- a/packages/vue-flicking/demo/plugins/Autoplay.vue
+++ b/packages/vue-flicking/demo/plugins/Autoplay.vue
@@ -39,7 +39,7 @@ export default class AutoPlayDemo extends Vue {
   plugins = [new Fade(), new AutoPlay(2000, "NEXT")];
   code0 = `<!--
   import { Fade, AutoPlay } from "@egjs/flicking-plugins";
-  plugins = [new Fade(), new AutoPlay(2000, "NEXT")];
+  const plugins = [new Fade(), new AutoPlay(2000, "NEXT")];
 -->
 <flicking
   class="flicking"
--- a/packages/vue-flicking/demo/plugins/Autoplay.vue
+++ b/packages/vue-flicking/demo/plugins/Autoplay.vue
@@ -37,7 +37,11 @@ export default class AutoPlayDemo extends Vue {
   }
 
   plugins = [new Fade(), new AutoPlay(2000, "NEXT")];
-  code0 = `<flicking
+  code0 = `<!--
+  import { Fade, AutoPlay } from "@egjs/flicking-plugins";
+  plugins = [new Fade(), new AutoPlay(2000, "NEXT")];
+-->
+<flicking
   class="flicking"
   :options="{ circular: true, gap: 10, duration: 500 }"
   :plugins="this.plugins"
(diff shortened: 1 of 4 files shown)


packages/vue-flicking/demo/plugins/Autoplay.vue
```diff
@@ -37,7 +37,11 @@ export default class AutoPlayDemo extends Vue {
   }
 
   plugins = [new Fade(), new AutoPlay(2000, "NEXT")];
-  code0 = `<flicking
+  code0 = `<!--
+  import { Fade, AutoPlay } from "@egjs/flicking-plugins";
+  plugins = [new Fade(), new AutoPlay(2000, "NEXT")];
```
daybrush: @WoodNeck

It seems'd better add a `const`.

packages/vue-flicking/src/Flicking.ts
```diff
@@ -149,7 +146,7 @@ export default class Flicking extends Vue {
 
   private $_getPanels(h: CreateElement) {
     const flicking = this.$_nativeFlicking;
-    const wholeSlots = this.$slots.default;
+    const wholeSlots = this.$_getSlots();
 
     if (!wholeSlots) {
```
daybrush: @WoodNeck


It seems to be true because wholeSlots is always an array. How about using length in this case?

packages/vue-flicking/src/Flicking.ts
```diff
@@ -149,7 +146,7 @@ export default class Flicking extends Vue {
 
   private $_getPanels(h: CreateElement) {
     const flicking = this.$_nativeFlicking;
-    const wholeSlots = this.$slots.default;
+    const wholeSlots = this.$_getSlots();
```
daybrush: @WoodNeck

It seems to be true because wholeSlots is always an array. How about using length in this case?

WoodNeck: Thanks for checking! I'll fix that.

packages/vue-flicking/demo/plugins/Autoplay.vue
WoodNeck: I'll fix that!

PR #334, opened 2019-12-17
fix: prevent scrolling when resize

### LINKED ISSUE #333: Flicking can be scrolled to top when address bar in mobile browser appear/disappears.
## Description
When the address bar in the mobile browser appear or disappears, the `resize` event is triggered.
When Flicking has `autoResize` option, Flicking will attempt to calculate viewport size based on new panel size.
And in that sequence, Flicking will set viewport height to ""(none) to calculate viewport-dependent panel size correctly. (see #201)
As all panel has `position:absolute`, viewport's height will be once 0, so it will be scrolled to -y based on viewport's size.

1. Make Flicking with sufficient height(horizontal: true), and option `autoResize: true`
2. On the mobile device, scroll enough to make address bar disappear so resize event can be triggered.

WoodNeck:
## Issue
#333

## Details
This fixes #333

### DIFF
--- a/src/components/Viewport.ts
+++ b/src/components/Viewport.ts
@@ -597,8 +597,10 @@ export default class Viewport {
       const viewportSize = `${sizeToApply}px`;
       if (horizontal) {
         viewportStyle.height = viewportSize;
+        state.cachedBbox!.height = sizeToApply;
       } else if (!horizontal) {
         viewportStyle.width = viewportSize;
+        state.cachedBbox!.width = sizeToApply;
       }
     }
   }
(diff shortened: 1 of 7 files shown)


src/components/Viewport.ts
```diff
@@ -597,8 +597,10 @@ export default class Viewport {
       const viewportSize = `${sizeToApply}px`;
       if (horizontal) {
         viewportStyle.height = viewportSize;
+        state.cachedBbox!.height = sizeToApply;
       } else if (!horizontal) {
```

Is it just else?

WoodNeck: That definitely is. My mistake.

PR #344, opened 2020-03-04
fix: touchmove not triggered when renderOnlyVisible option is enabled

### LINKED ISSUE #342: Using renderOnlyVisible option make panning stop
## Description
This happens only on mobile devices with touch events enabled.
As the `touchmove` / `touchend` event can be triggered when `touchstart` event's target element is not removed, renderOnlyVisible can make Flicking to stop.

1. Create Flicking instance with renderOnlyVisible option
2. Touch on one panel, and pan to make it disappear on the screen.

WoodNeck:
## Issue
#342

## Details
This fixes #342

### DIFF
--- a/src/components/Panel.ts
+++ b/src/components/Panel.ts
@@ -434,7 +434,7 @@ class Panel implements FlickingPanel {
   public removeElement(): void {
     if (!this.viewport.options.renderExternal) {
       const element = this.element;
-      element.parentNode!.removeChild(element);
+      element.parentNode && element.parentNode.removeChild(element);
     }
 
     // Do the same thing for clones
@@ -447,7 +447,7 @@ class Panel implements FlickingPanel {
     const options = this.viewport.options;
     const removingPanels = this.clonedPanels.splice(start);
 
-    if (!options.renderExternal && !options.renderOnlyVisible) {
+    if (!options.renderExternal) {
       removingPanels.forEach(panel => {
         panel.removeElement();
       });
--- a/src/components/Viewport.ts
+++ b/src/components/Viewport.ts
@@ -53,10 +53,6 @@ export default class Viewport {
     };
     infiniteThreshold: number;
     checkedIndexes: Array<[number, number]>;
-    visibleIndex: {
-      min: number;
-      max: number;
-    };
     isAdaptiveCached: boolean;
     isViewportGiven: boolean;
     isCameraGiven: boolean;
@@ -82,10 +78,6 @@ export default class Viewport {
       scrollArea: {
         prev: 0,
         next: 0,
-      },
-      visibleIndex: {
-        min: NaN,
-        max: NaN,
       },
       translate: TRANSFORM,
       infiniteThreshold: 0,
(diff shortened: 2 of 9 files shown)


src/components/Panel.ts
```diff
@@ -434,7 +434,7 @@ class Panel implements FlickingPanel {
   public removeElement(): void {
     if (!this.viewport.options.renderExternal) {
       const element = this.element;
-      element.parentNode!.removeChild(element);
+      element.parentNode && element.parentNode.removeChild(element);
```
mixed: The old code also looks good, why did it change? :)

src/components/Viewport.ts
```diff
@@ -1770,68 +1783,67 @@ export default class Viewport {
...
+      // Visible panels not changed
+      return;
+    }
 
-      if (newVisiblePanels.length > 0) {
-        const firstVisiblePanelPos = newVisiblePanels[0].getPosition();
-        state.positionOffset = firstVisiblePanelPos;
-      }
+    if (currentState.holding) {
+      newVisiblePanels.push(...removedPanels);
```
mixed: WOW. I learned expression. Thanks.

src/components/Viewport.ts
```diff
@@ -709,12 +724,58 @@ export default class Viewport {
...
+        }
+        return visiblePanels;
+      };
+
+      const panelCount = this.panelManager.getPanelCount();
+      const getAbsIndex = (panel: Panel) => panel.getIndex() + (panel.getCloneIndex() + 1) * panelCount;
+      const nextPanels = getVisiblePanels(basePanel, getNextPanel, isOutOfBoundNext);
+      const prevPanels = getVisiblePanels(basePanel, getPrevPanel, isOutOfBoundPrev);
+
+      return [basePanel, ...nextPanels, ...prevPanels].sort((panel1, panel2) => getAbsIndex(panel1) - getAbsIndex(panel2));
```

I've been thinking about this code and it looks like the code below is not a condition?
```js
return allPanels.filter(panel => {
        const outsetProgress = panel.getOutsetProgress();
        return outsetProgress > -1 && outsetProgress < 1;
      });
```

src/components/Viewport.ts
```diff
@@ -1894,9 +1906,9 @@ export default class Viewport {
     };
 
     // Stopped on first cloned first panel
-    if (lastPanelOfPrevDir.getIndex() === 0 && lastPanelOfPrevDir.getCloneIndex() === 0) {
-      newVisibleIndex.min = allPanelCount;
-    }
+    // if (lastPanelOfPrevDir.getIndex() === 0 && lastPanelOfPrevDir.getCloneIndex() === 0) {
```

Is this code now unnecessary?

src/components/Viewport.ts
WoodNeck: You're right. That should work for the non-renderOnlyVisible cases.
I'll see that works.

src/components/Viewport.ts
WoodNeck: Yes not they're unnecessary. I've just forgot to remove them.

src/components/Panel.ts
WoodNeck: There can be a case that elements can be removed without parentNode.
It's when `remove()` is called by that removed panel is part of visible panels so that it could be removed twice. That code prevents such case.

src/components/Viewport.ts
WoodNeck: I figured why. That's basically more efficient to do that.
Filtering all panels can be unnecessary when cases like there're plenty of panels but only 3 panels are visible.

PR #377, opened 2020-09-21
fix: wrong camera position after resizing

### LINKED ISSUE #376: Camera position can be wrong when resized with "freeScroll" and "bound"
## Description
Camera position can be wrong when resized with the option "freeScroll" and "bound".
This is probably due to the wrong panelMaintainRatio & nearestPanel value on Viewport's `updateCameraPosition` method.
See https://github.com/naver/egjs-flicking/blob/master/src/components/Viewport.ts#L606

https://codepen.io/woodneck/pen/LYNBYRX
- With forking the above example, remove where it compares camera position with scroll area limit then scroll.
- After initializing Flicking, click the second panel
- Make Flicking resized by changing the browser window size

WoodNeck:
## Issue
#376

## Details
This fixes #376
This also adds test cases for when all panels have 0 widths.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "@egjs/flicking",
-	"version": "3.5.0",
+	"version": "3.5.0-snapshot",
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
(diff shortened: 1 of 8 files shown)


mixed (approved):
LGTM

daybrush (approved):
LGTM

PR #383, opened 2020-10-28
fix: renderOnlyVisible not correctly positioning panels on the first render

### LINKED ISSUE #382: Panels are not positioned correctly with renderOnlyVisible option
## Description
Panels are not positioned correctly with renderOnlyVisible option when if all panels can be shown on the full viewport.

1. Enable "renderOnlyVisible" option
2. Place panels to not exceed viewport size, make sure any of the panels are hidden.

WoodNeck:
## Issue
#382

## Details
This fixes #382

### DIFF
--- a/src/components/Viewport.ts
+++ b/src/components/Viewport.ts
@@ -259,6 +259,7 @@ export default class Viewport {
     this.updateAdaptiveSize();
     this.updateScrollArea();
     this.updateClonePanels();
+    this.updateVisiblePanelPositions();
     this.updateCameraPosition();
     this.updatePlugins();
   }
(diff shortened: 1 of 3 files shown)


src/components/Viewport.ts
```diff
@@ -1086,6 +1086,10 @@ export default class Viewport {
     this.setDefaultPanel();
     this.resize();
     this.moveToDefaultPanel();
+    if (this.options.renderOnlyVisible) {
```
daybrush, line 1089: @WoodNeck 
this part is only for the first rendering.  I know that css is not injected even if resized.

mixed (approved):

daybrush (approved):
LGTM

PR #390, opened 2020-11-24
fix: wrong rendering indexes after prepending panels

### LINKED ISSUE #389: Prepending panels with setState can break carousel with renderOnlyVisible option
## Description
Flicking's `getRenderingIndexes` can return the wrong result with non-circular & renderOnlyVisible option enabled.
This issue is reported on the React framework but can occur across all frameworks.

```tsx
<Flicking
  className="flicking flicking0"
  gap={10}
  defaultIndex={1}
  renderOnlyVisible={true}
  onMoveEnd={(e) => {
    if (e.index === 0) {
      this.setState((state) => {
        const start = state.list0[0] || 0;
        return {
          list0: [
            start - 10,
            start - 9,
            start - 8,
            start - 7,
            start - 6,
            start - 5,
            start - 4,
            start - 3,
            start - 2,
            start - 1,
            ...state.list0
          ]
        };
      });
    }
  }}
>
  {this.state.list0.map((num) => {
    return (
      <div
        key={num}
        className={"infinite infinite" + (Math.abs(num) % 5)}
      >
        {num}
      </div>
    );
  })}
</Flicking>
```

WoodNeck:
## Issue
#389 

## Details
This fixes #389

### DIFF
--- a/src/Flicking.ts
+++ b/src/Flicking.ts
@@ -631,6 +631,7 @@ class Flicking extends Component<{
    */
   public getRenderingIndexes(diffResult: DiffResult<any>): number[] {
     const viewport = this.viewport;
+    const isCircular = this.options.circular;
     const visiblePanels = viewport.getVisiblePanels();
     const maintained = diffResult.maintained.reduce((values: {[key: number]: number}, [before, after]) => {
       values[before] = after;
@@ -651,7 +652,9 @@ class Flicking extends Component<{
         const cloneIndex = Math.floor(val / prevPanelCount);
         const changedIndex = maintained[val % prevPanelCount];
 
-        return changedIndex + panelCount * cloneIndex;
+        return isCircular
+          ? changedIndex + panelCount * cloneIndex
+          : changedIndex;
       });
 
     const renderingPanels = [...visibleIndexes, ...added];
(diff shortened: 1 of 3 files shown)


mixed (approved):
LGTM

src/Flicking.ts
```diff
@@ -651,7 +652,9 @@ class Flicking extends Component<{
         const cloneIndex = Math.floor(val / prevPanelCount);
```
daybrush, line 652: `visibleIndexes` is like the latest index. Is it right to use `prevPanelCount`?

If `circular` is used, the `cloneIndex` will be very large.

daybrush (approved):

src/Flicking.ts
WoodNeck: `val` is using an absolute index based on previous panel count, see line 645

daybrush: In the test, beforeSync was called first, so I thought that the indexes of visiblePanels were all changed.

is correct that cloneIndex and index do not change even after beforeSync?

WoodNeck: Well, that should be handled differently. I'll send a PR tomorrow.

PR #391, opened 2020-11-26
fix: wrong rendering indexes with renderOnlyVisible: true & circular: true

WoodNeck:
## Issue
#389

## Details
This fixes the additional issue of #389 that adding `circular: true` option can reproduce the issue.

### DIFF
--- a/src/Flicking.ts
+++ b/src/Flicking.ts
@@ -624,9 +624,9 @@ class Flicking extends Component<{
 
   /**
    * Get indexes to render. Should be used with `renderOnlyVisible` option.
-   * `beforeSync` should be called before this method for a correct result
+   * `beforeSync` should be called before this method for a correct result.
    * @private
-   * @ko 렌더링이 필요한 인덱스들을 반환한다. `renderOnlyVisible` 옵션과 함께 사용해야 한다.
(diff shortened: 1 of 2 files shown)


src/Flicking.ts
```diff
@@ -624,14 +624,14 @@ class Flicking extends Component<{
 
   /**
    * Get indexes to render. Should be used with `renderOnlyVisible` option.
+   * `beforeSync` should be called before this method for a correct result
```
mixed, line 627: I think I should also add Korean too.

mixed (approved):
LGTM

src/Flicking.ts
```diff
@@ -642,20 +642,11 @@ class Flicking extends Component<{
...
-      .map(val => {
-        const cloneIndex = Math.floor(val / prevPanelCount);
-        const changedIndex = maintained[val % prevPanelCount];
-
-        return isCircular
-          ? changedIndex + panelCount * cloneIndex
-          : changedIndex;
-      });
+    const visibleIndexes = visiblePanels.map(panel => getPanelAbsIndex(panel))
+      .filter(val => maintained[val % prevPanelCount] != null);
```
daybrush, line 649: Maybe this should be filtered by `maintained(after)` and `panelCount`?

src/Flicking.ts
WoodNeck: Oh, I forgot. I'll add it.

daybrush (approved):
LGTM

PR #397, opened 2020-11-30
feat: add resizeOnImagesReady option

### LINKED ISSUE #396: Add the resizeOnImagesReady option
## Description
Panel sizes can be changed after the image is loaded, and as a carousel component often includes such images I suggest adding a `resizeOnImagesReady` option to resolve that kind of issue.
@egjs/imready can be used as a dependency of Flicking for this feature.

WoodNeck:
## Issue
#396

## Details
This closes #396


mixed (commented):
I have a few questions.
1. The path of the testcase seems to be set incorrectly. Is it correct?
2. You need to check from a little more edge case. 
    - For example, what does Flicking do when an image(or video) fails to load? And what if I want to load the image again?

src/consts.ts
```diff
@@ -62,6 +62,7 @@ export const DEFAULT_OPTIONS: Readonly<FlickingOptions> = {
   isConstantSize: false,
   renderOnlyVisible: false,
   renderExternal: false,
+  resizeOnImagesReady: false,
```
mixed, line 65: Are only images available? Isn't video possible?

src/Flicking.ts
```diff
@@ -828,9 +831,25 @@ class Flicking extends Component<{
...
   private listenResize(): void {
-    if (this.options.autoResize) {
+    const options = this.options;
+
+    if (options.autoResize) {
       window.addEventListener("resize", this.resize);
     }
+
+    if (options.resizeOnImagesReady) {
+      const imageReadyChecker = new ImReady();
```
mixed, line 841: Should you create ImReady every time?

src/Flicking.ts
```diff
@@ -828,9 +831,25 @@ class Flicking extends Component<{
...
 
   private listenResize(): void {
-    if (this.options.autoResize) {
+    const options = this.options;
+
+    if (options.autoResize) {
       window.addEventListener("resize", this.resize);
     }
+
+    if (options.resizeOnImagesReady) {
```
daybrush, line 840: When destroying before becoming ready, imready should also be removed.

WoodNeck:
@mixed 
I've revised some specs based on your review, please check it again :)

@daybrush
I've found out that `preReady` is not triggered when there're images with `loading="lazy"` only, which breaks my tests.
Is that a real spec of imready?

src/Flicking.ts
```diff
@@ -559,7 +566,11 @@ class Flicking extends Component<{
     const parsedElements = parseElement(element);
 
     const insertingIndex = Math.max(viewport.panelManager.getRange().min - parsedElements.length, 0);
-    return viewport.insert(insertingIndex, parsedElements);
+    const prependedPanels = viewport.insert(insertingIndex, parsedElements);
+
+    this.contentsReadyChecker.check(prependedPanels.map(panel => panel.getElement()));
```
mixed, line 571: If `resizeOnContentsReady` is false, isn't `this.contentsReadyChecker` undefined? I think that You need to check everywhere you have used `this.contentsReadyChecker` . And the statement is repeated. How about making it a method?

mixed (commented):
Thank you. What I suggested seems to have worked well. And If the path is correct, then the coverage is reduced by 60%, so confirmation is required.

WoodNeck:
@mixed 
Well, you can ignore the coverage. I left `only` in the test by mistake. so it will be increased after the test is successful.

src/Flicking.ts
WoodNeck: Thanks, I've fixed it.

mixed (approved):
LGTM!

src/Flicking.ts
```diff
@@ -495,6 +501,7 @@ class Flicking extends Component<{
     }
 
     this.viewport.destroy(option);
+    this.contentsReadyChecker?.off();
```
daybrush, line 504: It is better to use destroy rather than off.

WoodNeck: I've fixed it, can you check what I've written above?

daybrush:
@WoodNeck 

In imready, all the test codes related to loading have passed, and it works well in UI tests.

test/unit/options.spec.ts
```diff
@@ -1458,4 +1456,241 @@ describe("Initialization", () => {
...
+        Object.defineProperty(imgEl, "complete", { value: false });
+
+        panelEl.appendChild(imgEl);
+        return panelEl;
+      });
+
+      flicking.resize = resizeSpy;
+
+      // When
+      flicking.append(newPanelElements);
```
daybrush, line 1579: @WoodNeck 

Do you check imready for the appended element?

daybrush:
@WoodNeck 

Would you like to update to version 1.1.0?

WoodNeck:
@daybrush 
I'm using v1.1.0 in this PR

daybrush:
@WoodNeck 

Have you ever changed the timing?
After checking, setTimeout does not occur.

daybrush (approved):
LGTM

PR #408, opened 2021-01-14
fix: next/prev/moveTo not work after calling disableInput while holding the panel

### LINKED ISSUE #407: Methods won't work after calling `disableInput` while holding the panel
## Description
Methods won't work after calling `disableInput` while holding the panel.
This is because the internal state is not changing after calling `disableInput`.

Call disableInput on touchstart, then call next/prev/moveTo after.

WoodNeck:
## Issue
#407 

## Details
This fixes #407

### DIFF
--- a/src/components/Viewport.ts
+++ b/src/components/Viewport.ts
@@ -23,7 +23,7 @@ export default class Viewport {
 
   private flicking: Flicking;
   private axes: Axes;
-  private panInput: PanInput;
+  private panInput: PanInput | null;
 
   private viewportElement: HTMLElement;
   private cameraElement: HTMLElement;
@@ -384,11 +384,26 @@ export default class Viewport {
   }
 
   public enable(): void {
-    this.panInput.enable();
+    if (!this.panInput) {
+      const options = this.options;
+      this.panInput = new PanInput(this.viewportElement, {
+        inputType: options.inputType,
+        thresholdAngle: options.thresholdAngle,
+        iOSEdgeSwipeThreshold: options.iOSEdgeSwipeThreshold,
+        scale: options.horizontal ? [-1, 0] : [0, -1],
+      });
+
+      this.axes.connect(options.horizontal ? ["flick", ""] : ["", "flick"], this.panInput);
+    }
   }
 
   public disable(): void {
-    this.panInput.disable();
+    if (this.panInput) {
+      this.panInput.destroy();
+      this.panInput = null;
+
+      this.stateMachine.transitTo(STATE_TYPE.IDLE);
+    }
   }
 
   public insert(index: number, element: ElementLike | ElementLike[]): FlickingPanel[] {
@@ -674,7 +689,7 @@ export default class Viewport {
     }
 
     this.axes.destroy();
-    this.panInput.destroy();
+    this.panInput && this.panInput.destroy();
 
     originalPanels.forEach(panel => { panel.destroy(option); });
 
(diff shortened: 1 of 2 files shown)


src/components/Viewport.ts
```diff
@@ -674,7 +689,7 @@ export default class Viewport {
     }
 
     this.axes.destroy();
-    this.panInput.destroy();
+    this.panInput && this.panInput.destroy();
```
mixed, line 692: nit
```diff
- this.panInput && this.panInput.destroy();
+ this.panInput!.destroy();
```

mixed (approved):

src/components/Viewport.ts
WoodNeck: I thought Flicking was still using Typescript v2.8, I'll fix that.

src/components/Viewport.ts
```diff
@@ -384,11 +384,26 @@ export default class Viewport {
...
+        scale: options.horizontal ? [-1, 0] : [0, -1],
+      });
+
+      this.axes.connect(options.horizontal ? ["flick", ""] : ["", "flick"], this.panInput);
+    }
   }
 
   public disable(): void {
-    this.panInput.disable();
+    if (this.panInput) {
```
daybrush, line 401: @WoodNeck 

It probably overlaps with the init code.

src/components/Viewport.ts
```diff
@@ -384,11 +384,26 @@ export default class Viewport {
...
+      this.panInput.destroy();
+      this.panInput = null;
```
daybrush, line 403: Is the original paninput disconnected?

src/components/Viewport.ts
```diff
@@ -384,11 +384,26 @@ export default class Viewport {
...
+    if (!this.panInput) {
+      const options = this.options;
+      this.panInput = new PanInput(this.viewportElement, {
+        inputType: options.inputType,
+        thresholdAngle: options.thresholdAngle,
+        iOSEdgeSwipeThreshold: options.iOSEdgeSwipeThreshold,
+        scale: options.horizontal ? [-1, 0] : [0, -1],
+      });
+
+      this.axes.connect(options.horizontal ? ["flick", ""] : ["", "flick"], this.panInput);
```
daybrush, line 396: @WoodNeck

It probably overlaps with the init code.

src/components/Viewport.ts
WoodNeck: Yes it calls `disconnect()` on destroy.
image

daybrush: @WoodNeck 

I mean, it is said to `axes.disconnect`

WoodNeck: Looks like `axes.disconnect` also calls `disconnect` of a inputType, so I guess it's okay?
image

src/components/Viewport.ts
WoodNeck: Thanks for checking, I've updated my code

src/components/Viewport.ts
daybrush: Um... it needs to call the destroy method of paninput and then the disconnect method.

WoodNeck: Why should it have to be?

daybrush: Well. Should the axes be connected, shouldn't I disconnect? Even if the axes don't call disconnect, there seems to be no problem.

daybrush (approved):
LGTM

src/components/Viewport.ts
WoodNeck: I mean, as the PanInput calls `disconnect()` while calling `destroy()` why should it have to be called again?

Issue #234, opened 2019-05-29
Support Preact framework


daybrush:
## Description

Support Preact framework.
Usage is like flicking..

PR #411, opened 2021-01-19
feat: add preact-flicking package

daybrush:
## Issue

#234


mixed (approved):
Welcome to Flicking for @preactjs

packages/preact-flicking/src/preact-flicking/Flicking.js
```diff
@@ -0,0 +1,3 @@
+import Flicking from "@egjs/react-flicking";
+
+export default Flicking;
```
WoodNeck, line 3: I think there could be a reason to make this file in `js` format, can I ask why?

WoodNeck (approved):
LGTM

packages/preact-flicking/src/preact-flicking/Flicking.js
daybrush: @WoodNeck 

One of the ways to convert from react to preact is to remove the react type.
So, I created Flicking.js and Flicking.d.ts.  React was removed from Flicking.d.ts and changed to preact.

Issue #409, opened 2021-01-14
We're planning the v4!


WoodNeck:
We're actually preparing for version 4.
While maintaining the v3 version, we felt the need to upgrade the major version due to many issues that can't be solved.

These are common issues that can't be fixed until we're upgrading to v4.

## Cloning of panel elements
Because the actual panel elements are being cloned, there were many issues when there're many elements inside a single panel or an iframe advertisement exists on the panel.
The only good case of actually creating cloned elements we can found was this demo: https://codepen.io/woodneck/pen/agoWOj
By that, we're planning to remove cloning panel elements in v4 and prevent `circular` option when there're not enough panels exist.

## Empty panel indexes can exist
For example, when there're panels at indexes 0 and 2 exist, the panel at index 1 may not exist.
We tried to apply `needPanel` events of the `infinite` option into a wider range of cases, but we think they only made it difficult to use.
So we're gonna remove the `replace` method of the Flicking and replace it with `insert` instead.

## Panel CSS
We're using the `position: absolute` and `left/top` property of the panel elements to place them.
But as it requires the fixed size of either the viewport or panel. and as it also prevents the user from naturally implementing animation with the panel's height or width property, we're going to use another CSS property to solve these issues.
Also, for the SSR(Server-Side Rendering) applications we're planning to distribute the CSS files instead of setting them with the script.

## HTML Structure
Currently, we're adding 2 additional HTML elements via script. But for a more consistent layout between applying the Flicking script or not, we're going to have users add their own elements on their markup.

## Other changes that we're planning
- `needPanel` event will be triggered also for the "prev" direction, event if when it's on the index 0.
- We'll use `useOffset: true` as a default value.
- `direction` property of the `change` event will be changed to indicate the actual direction of the panel index change.

----

And of course, we'll gonna solve the current open issues on version 4.
If you think there're other changes needed in the Flicking functions or interfaces, please leave a comment below :)

PR #412, opened 2021-01-22
feat: design basic structure of Flicking v4

WoodNeck:
## Issue
#409


mixed (commented):
LGTM
I reviewed repetitive code and switch refactor.

daybrush:
Using promises in v4's internal code doesn't support IE?

WoodNeck:
> Using promises in v4's internal code doesn't support IE?

That's right, now it requires the polyfill for environments like IE.

src/camera/Camera.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  public getElement() { return this._el; }
+  public getPosition() { return this._position; }
+  public getAlignPosition() { return this._alignPos; }
+  public getRange() { return this._range; }
+
+  public getPanelBelow(): Panel | null {
+    const flicking = this._flicking;
+
+    if (!flicking) {
```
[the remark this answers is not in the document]
WoodNeck: That's a good idea, I'll try that.

src/control/StateMachine.ts
```diff
@@ -0,0 +1,82 @@
...
+        break;
+      case AXES_EVENT.RELEASE:
+        currentState.onRelease(ctx);
+        break;
+      case AXES_EVENT.ANIMATION_END:
+        currentState.onAnimationEnd(ctx);
+        break;
+      case AXES_EVENT.FINISH:
+        currentState.onFinish(ctx);
+        break;
```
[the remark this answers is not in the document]
WoodNeck: Well if it's not a file size problem, I personally prefer that style as I think it can explain the meaning of the code better.
It is also a more typescript-friendly approach.

src/core/Viewport.ts
```diff
@@ -0,0 +1,76 @@
...
+    height: number | string;
+  }) {
+    const el = this._el;
+    if (width != null) {
+      if (typeof width === "string") {
+        el.style.width = width;
+      } else {
+        el.style.width = `${width}px`;
+      }
+    }
```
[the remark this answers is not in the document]
WoodNeck: That's cool, I'll try that.

src/const/external.ts
```diff
@@ -0,0 +1,65 @@
...
+  NEXT: "next"
+};
+
+export const DIRECTION: {
+  PREV: "PREV";
+  NEXT: "NEXT";
+} = {
+  PREV: "PREV",
+  NEXT: "NEXT"
+};
```
daybrush, line 65: Add `as const` to the constant. There is no need to set the type separately.

WoodNeck: This really helps, thanks

daybrush (approved):
LGTM

src/camera/Camera.ts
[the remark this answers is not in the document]
WoodNeck: Well I found out that TS decorators use `Object.getOwnPropertyDescriptor`, which doesn't support IE 8.
Like the case of `getter/setter`, decorators can't be used for that reason.

src/control/StateMachine.ts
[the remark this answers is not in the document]
mixed: In general, in the case of switch statements, when the type increases, the probability of generating an error is high, so it is better to improve. However, in the current case, it doesn't seem to be bad even now, since it actually converts the old code. You may find it helpful later by referring to the article below.

- https://stackoverflow.com/questions/54079960/refactoring-switch-statement-typescript
- https://medium.com/itchis-coffee-shop/how-to-refactor-a-switch-statement-with-object-composition-in-angular-ba2bc308b7d2

mixed (approved):
🙌🙌🙌🙌🙌

src/control/StateMachine.ts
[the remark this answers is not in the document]
WoodNeck: I don't agree with that. Not all switch statements are bad.
The second link that you gave tells that switch statements are bad for the following two reasons.

### 1. It's error-prone, because related codes should be combined to each case.

The author of that article insists that **Object Composition & Factory Method** is the solution to bypass switch. 
But as you can see I'm using the exact same thing in my code. `transitTo` is the factory method itself, and I'm using object composition in each state(yes, this is also a state pattern).
And for the `fire`, it just propagates the Axes event to the state. This is also a case of object composition by using state.
I said my code is more typescript-friendly because changing code like that will make the state call `any` typed method.
image
And also, calling methods like that is more error-prone as just one typo can create all kinds of errors, and it can't be statically caught.

### 2. It's time-comsuming, as adding more cases will take some time
I'm half-agree with this. I agree that adding more cases will take more time, but in this case, we're selectively choosing Axes event to listen to, which is very limited and also have to be precise.
And as we're using Typescript, we can easily notice the missing cases. Like, I'm not using the `default` case in my code but still have no ts/lint error because I'm handling all the cases given. Adding more types will emerge the static error, so we can easily handle that.
image
> Simply, adding more enum value shows error on my IDE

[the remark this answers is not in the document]
mixed: I agree on some parts. So first of all, I don't want to say that I have to do that. The point I want to say is that in the case of a switch statement, the points that change according to the added state increase. For example, when the type of state.SUPER increases in the swtich statement, the point that needs to be modified is the part of adding the switch statement and the corresponding method. If the switch is changed to the example above, it is said that it is good because the point to be concerned with is reduced from 2 to 1 because only objects that match the state are created. This is because there is a high probability of generating errors if the number of parts that are changed according to some change in development increases.

However, the reason I didn't think this part was very important is that it is already stable code, so the state change is unlikely, so it is good even now. :)

src/Flicking.ts
```diff
@@ -5,560 +5,586 @@
...
+ * @param {boolean} [options.resizeOnContentsReady=false] Whether to resize the Flicking after the image/video elements inside viewport are ready.<br/>Use this property to prevent wrong Flicking layout caused by dynamic image / video sizes.<ko>Flicking 내부의 이미지 / 비디오 엘리먼트들이 전부 로드되었을 때 Flicking의 크기를 재계산하기 위한 옵션.<br/>이미지 / 비디오 크기가 고정 크기가 아닐 경우 사용하여 레이아웃이 잘못되는 것을 방지할 수 있다.</ko>
+ * @param {boolean} [options.collectStatistics=true] Whether to collect statistics on how you are using `Flicking`. These statistical data do not contain any personal information and are used only as a basis for the development of a user-friendly product.<ko>어떻게 `Flicking`을 사용하고 있는지에 대한 통계 수집 여부를 나타낸다. 이 통계자료는 개인정보를 포함하고 있지 않으며 오직 사용자 친화적인 제품으로 발전시키기 위한 근거자료로서 활용한다.</ko>
+ */
+export interface FlickingOption {
```
daybrush, line 82: The singular and plural are mixed in the interface name.

And there is some overlap with types.ts.

Is Flicking.ts a temporary interface?

WoodNeck: types.ts is an old file that I will eventually remove so you can ignore that.
I'm trying to use only singular form overall so if there's an interface name with plural that's my miatake.

daybrush: Is the variable name going to be changed to singular as well? ex) `options`

WoodNeck: I'll use plural only when if variables/class members/argument type is an array type.
Except for the previous option names like `inputType` because of compatibility.

daybrush: The feeling of FlickingOption is like an enum name.

When I look at an object that uses that singular type, I think it's an individual rather than a group.

Wouldn't it be better to use group nouns if you wanted to use the object singularly?

WoodNeck: I can agree with that, as we commonly use the term "options" for option collections, and use "option" for a single option.
So, I'll gonna change collection-like interfaces like `option` to `options` and `event` to `events`.
But you know, we don't have any naming convention guidelines for naming those things.
I totally don't mind changing those namings before releasing v4, but we need more solid reasons to change those.
Like, your perspective can be very personal by looking at other types, what about `FlickingStatus` or `OriginalStyle`? Do they have to be `FlickingStatuses` and `OriginalStyles`?
The same things can be applied to our other repositories, like `AnimationParam` of Axes and `IDOMRendererSize` of InfiniteGrid.

PR #414, opened 2021-01-29
feat: panel add/remove and needPanel/visibleChange

WoodNeck:
## Issue
#409 

There're still lots of tests to be migrated..


src/control/states/DraggingState.ts
```diff
@@ -45,20 +46,19 @@ class DraggingState extends State {
...
     transitTo(STATE_TYPE.ANIMATING);
 
-    controller.setReleaseContext(axesEvent);
-    control.moveToPosition(axesEvent.destPos.flick, Math.max(axesEvent.duration, flicking.getDuration()));
-    controller.setReleaseContext(null);
+    const control = flicking.control;
+    const position = axesEvent.destPos.flick;
+    const duration = Math.max(axesEvent.duration, flicking.duration);
+
+    void control.moveToPosition(position, duration, axesEvent);
```
mixed, line 61: Why need to `void`?

mixed (approved):
LGTM

src/control/states/DraggingState.ts
WoodNeck: Because of the new eslint lint rule that requires `await` or `void` to all promise-based methods.
We can discuss whether to use it, but I think it's neat as it's more revealing that it's an asynchronous method.

src/Flicking.ts
```diff
@@ -649,10 +661,10 @@ class Flicking extends Component<FlickingEvent> {
     let type: MoveTypeStringOption | undefined;
     let moveTypeOptions: MoveTypeObjectOption | undefined;
 
-    if (isString(moveType) && moveType in DEFAULT_MOVE_TYPE_OPTIONS) {
+    if (typeof moveType === "string" && moveType in DEFAULT_MOVE_TYPE_OPTIONS) {
```
daybrush, line 664: I think we can solve the duplicate code, but is there a reason to remove isString?

WoodNeck: That's right, I think I've removed that because once while I'm working on my code there was a moment that `=== "string"` is used only once.
I'll revert that, thanks for checking :)

src/control/Control.ts
```diff
@@ -1,88 +1,106 @@
...
-    return this._controller!.animateTo(panel.getPosition(), duration)
-      .then(() => {
-        this._activeIndex = panel.getIndex();
-      });
+    const updateActivePanel = () => {
+      this._activePanel = panel;
+    };
+    const animate = () => this._controller.animateTo(panel.position, duration, axesEvent);
+
+    if (duration === 0) {
```
daybrush, line 98: When duration is 0, it seems to be a problem if the order of operation of animate => updateActivePanel is changed.

Since the duration is 0, there seems to be no problem right now, but if there is activePanel related code in animateTo, it will be a problem.

WoodNeck: That's an exceptional case.
As it's promise based method, it guarantees that it returns promise while updating the index synchronously.
And there won't be code that manipulates the active panel in AxesController, that's definitely not a role of it.

daybrush: You mean you have to compare if duration === 0?

I thought it was the same code because both return promises.

daybrush (approved):
LGTM

src/control/Control.ts
WoodNeck: That code is to match Axes's behavior.
Axes runs synchronously if the duration is 0, so it just updates the active index synchronously before the Promise is resolved.
It makes sure that the index is updated immediately after calling moveTo with duration 0.

```ts
if (duration > 0) {
  this.animateTo(movedPos, duration);
} else {
  this.em.triggerChange(movedPos);
  this.finish(false);
}
```
> Axes setTo

PR #415, opened 2021-02-01
Update README.md

kimduksoo:
to run below demo, need to install flicking-plugins also.

### DIFF
--- a/packages/vue-flicking/README.md
+++ b/packages/vue-flicking/README.md
@@ -35,6 +35,7 @@
 ## ⚙️ Installation
 ```sh
 npm install --save @egjs/vue-flicking
+npm install --save @egjs/flicking-plugins
 ```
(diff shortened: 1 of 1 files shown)


WoodNeck:
Hello @kimduksoo!
It is right that the `@egjs/flicking-plugins` also has to be installed to run the demo below,
but as it's not always necessary for all Flicking users, I think it is a bit dangerous to make them install it.
So I think it would be rather nice to create another section that describes how to use plugins with it.

kimduksoo:
@WoodNeck 
Yes, that's a good idea. Actually, I also create a plugin separately.

WoodNeck:
@kimduksoo 
Then I'll create another PR adding the notation about installing plugins as I've described above.
Thanks for the review :)

PR #416, opened 2021-02-09
feat: add more controls and cameras

WoodNeck:
## Issue
#409


src/camera/Camera.ts
```diff
@@ -149,9 +181,9 @@ abstract class Camera {
       return;
     }
 
-    const viewportSize = flicking.viewport.size;
+    const viewport = flicking.viewport;
     const cameraPosition = this._position;
-    const cameraSize = flicking.horizontal ? viewportSize.width : viewportSize.height;
+    const cameraSize = flicking.horizontal ? viewport.width : viewport.height;
```
mixed, line 186: The logic to find the size of the viewport is often seen. How about making it a function?
`flicking.horizontal ? viewport.width : viewport.height;`

src/camera/CircularCamera.ts
```diff
@@ -2,12 +2,186 @@
...
+    return super.lookAt(pos);
+  }
+
+  protected _applyTransform(): void {
+    const el = this._el;
+    const flicking = getFlickingAttached(this._flicking, "Camera");
+
+    el.style[this._transform] = flicking.horizontal
+      ? `translate(${-(this._position - this._alignPos + this._circularOffset)}px)`
+      : `translate(0, ${-(this._position - this._alignPos + this._circularOffset)}px)`;
```
mixed, line 166: This logic is also repeated.

src/control/Control.ts
```diff
@@ -16,6 +18,12 @@ abstract class Control {
   protected _controller: AxesController;
   protected _activePanel: Panel | null;
 
+  public get controller() { return this._controller; }
+  public get activeIndex() { return this._activePanel?.index ?? -1; }
```
mixed, line 22: Does the meaning of `-1` mean there is no active index? If so, I think you have to some add description.

mixed (approved):
🥇 🥇 🥇 🥇 🥇

src/camera/Camera.ts
WoodNeck: That's right, I'm considering adding it as a `camera.size` maybe

src/control/Control.ts
WoodNeck: As a flicking returns `-1` from `getIndex` when there are no panels that exist, that also returns `-1` when there is no active panel.
This is documented on the Flicking's `getIndex`, but not on the Control as there's no description on the Control yet.
I'll add the descriptions while adding the whole documentation of it :)

src/renderer/Renderer.ts
```diff
@@ -182,7 +205,7 @@ abstract class Renderer {
     const align = this._align;
 
     return typeof align === "object"
-      ? (align as { anchor: string | number }).anchor
+      ? (align as { panel: string | number }).panel
```
daybrush, line 208: Is there a type other than number, string, or object for align?

WoodNeck: Those three types that you've mentioned are only types that are used for the align option.

daybrush (approved):
LGTM

PR #418, opened 2021-02-11
fix: only call getCloneCount function when it exists

noelklein:
## Details
We've been getting errors on our website reported in New Relic:
```
this.flicking.getCloneCount is not a function:
183 | checkCloneCount() {
184 | if (!this.flicking) {
185 | return;
186 | }
187 |  
188 | const newCC = this.flicking.getCloneCount();
189 |  
190 | if (this.internalCloneCount === newCC) {
191 | return;
192 | }
```

This fix checks to make sure that the function actually exists.

### DIFF
--- a/packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
+++ b/packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
@@ -181,7 +181,7 @@ export class NgxFlickingComponent extends FlickingInterface
   }
 
   checkCloneCount() {
-    if (!this.flicking) {
+    if (!this.flicking || typeof(this.flicking.getCloneCount) !== 'function') {
       return;
     }
 


WoodNeck:
Hello @noelklein!
I wonder in what situation does getCloneCount not exist
Can you file an issue instead so I can debug it?

noelklein:
@WoodNeck I provided more info in this issue: https://github.com/naver/egjs-flicking/issues/419

WoodNeck:
> @WoodNeck I provided more info in this issue: #419

Thank you :) It might take a while but I'll check and reply to you.

noelklein:
@WoodNeck the fix I provided shouldn't hurt anything. Can we merge this PR as a temporary fix until you figured out the real problem?

WoodNeck:
@noelklein
It's really sorry but I'm not the only maintainer here, so I can't decide to merge certain commit alone.

I've just read your issue and found out that destroying Flicking before calling `checkCloneCount` can create that certain issue. So I think it would better to set `this.flicking` to null in ngOnDestroy to prevent other similar issue.

As it's also have to be confirmed by other maintainers and thus take a while, I suggest releasing a beta version instead so it won't conflict any other versions and you can take this patch before merging this PR.
If it's okay, I'll do that for you. It will be named like v3.5.3-beta or sth :)

noelklein:
@WoodNeck A beta version would be great, thank you!

WoodNeck:
@noelklein 
I've just released a beta version including this PR's changes.
You can install it by:
```sh
# This should install @egjs/ngx-flicking@3.4.3-beta.0
npm i --save @egjs/ngx-flicking@beta
```
Remember that this is a temporary fix that I've not tested so there can be any unexpected bug or issues.
Please feel free to contact me at issue #419 if there's any following bug/issue in the update :)

I'll close this PR and create a new one after I start working on this issue as I have to check other conditions related to where flicking does not exist. Anyway, thanks for the PR @noelklein!

mixed (approved):
Thank you for your PR!

daybrush (approved):

daybrush:
Thank you for PR

PR #420, opened 2021-02-16
feat: add VisibleRenderer

WoodNeck:
## Issue
#409


src/camera/Camera.ts
```diff
@@ -129,38 +177,53 @@ abstract class Camera {
...
+    this._anchors = [];
     this._needPanelTriggered = { prev: false, next: false };
   }
 
   protected _refreshVisiblePanels() {
     const flicking = getFlickingAttached(this._flicking, "Camera");
     const panels = flicking.renderer.panels;
 
-    const newVisiblePanels = panels.filter(panel => panel.isVisible());
+    const newVisiblePanels = panels.filter(panel => this.canSee(panel));
```
mixed, line 211: The verb `can` was a little awkward, but it's an appropriate naming. It also helped me read the code. :)

https://softwareengineering.stackexchange.com/questions/348649/naming-of-bool-methods-is-vs-can-vs#:~:text=According%20to%20the%20Microsoft%20naming,not%20what%20it%20can%20do.

src/camera/CircularCamera.ts
```diff
@@ -19,7 +20,68 @@ class CircularCamera extends Camera {
     };
   } = {};
 
-  public get circularEnabled() { return this._circularEnabled; }
+  public get controlParams() { return { range: this._range, position: this._position, circular: this._circularEnabled }; }
+
+  public getPrevAnchor(anchor: AnchorPoint): AnchorPoint | null {
```
mixed, line 25: When the return value of a method is an object, null is often returned when there is no object to return. How about creating an appropriate type and using it?

src/camera/CircularCamera.ts
```diff
@@ -148,20 +231,23 @@ class CircularCamera extends Camera {
...
     const flicking = getFlickingAttached(this._flicking, "Camera");
 
+    const actualPosition = this._position - this._alignPos - this._offset + this._circularOffset;
+
     el.style[this._transform] = flicking.horizontal
-      ? `translate(${-(this._position - this._alignPos + this._circularOffset)}px)`
-      : `translate(0, ${-(this._position - this._alignPos + this._circularOffset)}px)`;
+      ? `translate(${-actualPosition}px)`
+      : `translate(0, ${-actualPosition}px)`;
   }
```
mixed, line 251: It repeats with the code [here](https://github.com/naver/egjs-flicking/pull/420/files#diff-1b8ee851a75d0e755767cd49b1e3330cc8066793bf7660a7d6faa532b6541b48R302-R304). How about make a method?

src/core/Panel.ts
```diff
@@ -202,6 +150,21 @@ class Panel {
...
+
+  public decreaseOffset(val: number): this {
+    this._offset -= val;
+    return this;
+  }
+
+  public resetOffset(): this {
+    this._offset = 0;
+    return this;
+  }
```
mixed, line 166: How about making it with a common logic like below?

```ts
public increaseOffset(val: number): this {
    return _setOffset(val);;
}

public decreaseOffset(val: number): this {
    return _setOffset(-val);
}

public resetOffset(): this {
    return _setOffset(0);
}

private _setOffset(val:number): this {
    this._offset = val;
    return this;
}
```

src/renderer/VisibleRenderer.ts
```diff
@@ -4,8 +4,62 @@
...
+
+    const offsetGtZero = this._panels.filter(panel => panel.offset > 0);
+    const visibleOffsetLteZero = visiblePanels.filter(panel => panel.offset <= 0);
+
+    visibleSortedByActualPosition.forEach(panel => {
+      fragment.appendChild(panel.element);
+    });
+
+    // Remove remaining(removed) elements
+    while (cameraElement.firstChild) {
```
mixed, line 29: I think that`firstElementChild` seems logically more appropriate than `firstChild` with text nodes. And if you're going to delete all of them, it's a live DOM, so it looks better to process it with `innerHTML=""`. What do you think?

mixed (commented):
This PR was more readable because the code was more appropriately named. 👍

src/camera/Camera.ts
```diff
@@ -112,11 +102,69 @@ abstract class Camera {
...
+      }
+    }
+
+    return null;
+  }
+
+  public findNearestAnchor(position: number): AnchorPoint | null {
+    const anchors = this._anchors;
+
+    if (anchors.length <= 0) return null;
```
daybrush, line 128: It would be nice to make a constant called length.

src/camera/Camera.ts
WoodNeck: I named it to be natural like the case of the `if (camera.canSee(panel))` :)

src/camera/CircularCamera.ts
WoodNeck: What exactly is the appropriate type you mean?
Is it like `NullPoint`? or `type A = AnchorPoint | null`?

src/camera/CircularCamera.ts
WoodNeck: Okay, I'll refactor that.

src/core/Panel.ts
WoodNeck: Actually, this offset thing is a temporary implementation of the camera offset, and I'm wondering how to deal with it more properly.
As we can't modify the actual animation offset as we can't change Axes's animation value, I adopted this offset but I don't believe this is the perfect way for doing that.
So anyway, I just want to tell you that this offset can be removed somehow future.

src/renderer/VisibleRenderer.ts
WoodNeck: I believe I used removing all the `firstChild` method as it's the fastest way for doing that while I'm researching for Flicking v3.
- https://www.measurethat.net/Benchmarks/Show/10677/0/remove-all-children-from-dom-element-with-lastchild#latest_results_block
Also you're right that `firstElementChild` is more logically appropriate, but as Flicking shouldn't have text nodes inside I think it is also correct to remove all the child nodes.

src/camera/Camera.ts
WoodNeck: Okay, I'll do that.

src/renderer/VisibleRenderer.ts
```diff
@@ -4,8 +4,62 @@
...
+    const invisiblePrevPanels = this._panels.slice(0, visibleOffsetLteZero[0]?.index ?? 0);
+
+    const firstPanel = invisiblePrevPanels[0];
+    const lastPanel = invisiblePrevPanels[invisiblePrevPanels.length - 1];
+
+    // Calibrate offset created by circular
+    const firstGtPanel = offsetGtZero[0];
+    const lastGtPanel = offsetGtZero[offsetGtZero.length - 1];
+
+    const invisibleSize = this._calcPanelRangeSize(firstPanel, lastPanel)
```
daybrush, line 44: If the order of the elements(visibleSortedByActualPosition) is changed, is there any effect on invisibleSize?
If a panel with index 0 is visible, is it always 0?

WoodNeck: You're right, it is meant to be designed like that.

src/camera/CircularCamera.ts
mixed: This feeling? `type Nullable<T> = T | null;`

mixed (approved):

daybrush (approved):
LGTM

PR #422, opened 2021-02-17
fix: wrong camera position on setStatus

### LINKED ISSUE #421: Flicking's camera position can be wrong when getStatus is called from the non-zero index
## Description
Flicking's camera position can be wrong when getStatus is called from the non-zero index, which makes all panels to be invisible.

Call `getStatus()` from non-zero index and call `setStatus()` with that status.

WoodNeck:
## Issue
#421 

## Details
This fixes #421

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "@egjs/flicking",
-	"version": "3.7.1",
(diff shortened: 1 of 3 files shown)


mixed (approved):
LGTM

daybrush (approved):
LGTM

PR #424, opened 2021-02-22
feat: add element to select event

### LINKED ISSUE #417: React Flicking: Access of target DOM element in select event or prevent triggering of select event in some cases
## Description
It seems it is impossible access of target DOM element in `select` event.

I have a use case where panels contain several different interactive elements (for example, text, images and buttons).
When a user clicks on text or image inside a panel, `select` event should be processed.
But when a user clicks on a button, `select` event should be ignored (or bypassed) and only button’s click handler should be processed.
Unfortunately I’ve not found a way to achieve this behavior.
`Select` event handler is always triggered before button’s click event handler (even before capture handler).
There is no way to prevent triggering of `select` event depending of target DOM element.

Use the following code sandbox to check the issue: https://codesandbox.io/s/react-flicking-select-event-target-element-issue-8hop9
Click on an action button.
You will see in console that `select` event handler is run before button’s click handlers.
This is undesired and problematic in my case.

WoodNeck:
## Issue
#417 

## Details
This resolves #417

### DIFF
--- a/src/states/HoldingState.ts
+++ b/src/states/HoldingState.ts
@@ -5,7 +5,7 @@
 
 import State from "./State";
 import { STATE_TYPE, EVENTS, DIRECTION } from "../consts";
-import { FlickingContext } from "../types";
+import { FlickingContext, SelectEvent } from "../types";
 
 class HoldingState extends State {
   public readonly type = STATE_TYPE.HOLDING;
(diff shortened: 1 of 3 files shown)


mixed (approved):
LGTM

daybrush (approved):
LGTM

PR #425, opened 2021-02-22
fix: panel position is wrong after setStatus

### LINKED ISSUE #423: Wrong panel positions after calling setStatus
## Description
As Flicking currently doesn't save individual panel positions,
Camera's position can be wrong if the first panel's position is not at 0 when calling getStatus

Call getStatus while the first panel's position is not at 0
Then call setStatus with that status object.

WoodNeck:
## Issue
#423 

## Details
This fixes #423


mixed (approved):
LGTM

daybrush (approved):
LGTM

PR #431, opened 2021-03-10
feat: CFC for v4

WoodNeck:
## Issue
#409


packages/react-flicking/src/react-flicking/Flicking.tsx
```diff
@@ -1,188 +1,225 @@
...
   public componentWillUnmount() {
-    this.destroy({ preserveUI: true });
+    this._nativeFlicking.destroy();
   }
 
-  // private
-  private checkPlugins() {
-    const { list, added, removed, prevList } = this.pluginsDiffer.update(this.props.plugins!);
+  public shouldComponentUpdate(nextProps: this["props"]) {
+    const children = this._getChildren(nextProps.children);
```
daybrush, line 98: The code of shouldComponentUpdate and componentDidUpdate is too long. Is there a reason to handle it in the framework?

packages/react-flicking/src/react-flicking/Flicking.tsx
```diff
@@ -1,188 +1,225 @@
...
+      const cameraEl = this._cameraElement;
+
+      diffResult.added.forEach((panelIdx, elIdx) => {
+        const el = addedElements[elIdx];
+        const elNext = renderer.panels[panelIdx]
+          ? renderer.panels[panelIdx].element
+          : null;
+
+        if (el.nextElementSibling !== elNext) {
+          cameraEl.insertBefore(el, elNext);
```
daybrush, line 161: It is dangerous to add/remove/change DOM directly from the framework.

packages/react-flicking/src/react-flicking/Flicking.tsx
WoodNeck: Well my code was not sanitized so I've just updated them to use the same code.

packages/react-flicking/src/react-flicking/Flicking.tsx
WoodNeck: That was the code before I update the panel's resize logic, I forgot to remove that!

src/renderer/Renderer.ts
```diff
@@ -226,10 +226,17 @@ abstract class Renderer {
     const flicking = getFlickingAttached(this._flicking, "Renderer");
 
     const cameraElement = flicking.camera.element;
-    const cameraChilds = toArray(cameraElement.children);
+    const cameraChilds = toArray(cameraElement.childNodes);
```
mixed: Did you change it because of the range of browser support?

src/renderer/RawRenderer.ts
```diff
@@ -2,10 +2,202 @@
...
+    cameraElement.insertBefore(fragment, nextSiblingPanel);
+
+    // Resize the newly added panels
+    newPanels.forEach(panel => panel.resize());
+
+    // Update camera & control
+    camera.updateRange();
+    camera.updateAnchors();
+    camera.resetNeedPanelHistory();
+    control.updateInput();
```
mixed: This group seems to work as a combination. How about making it with a method?

src/renderer/RawRenderer.ts
```diff
@@ -2,10 +2,202 @@
...
+    const camRangeDiff = camera.rangeDiff;
+
+    this._relocatePanelElements(panels, null);
+
+    panels.forEach(panel => {
+      panel.increaseOffset(camRangeDiff);
+    });
+
+    return this;
+  }
```
mixed: nit) It seems to be similar logic. How about doing something like this?

```ts
private movePanelElements(panels: ElementPanel[], refElement, method){
    const flicking = getFlickingAttached(this._flicking, "Renderer");
    const camera = flicking.camera;
    const camRangeDiff = camera.rangeDiff;

    this._relocatePanelElements(panels, refElement);
    panels.forEach(panel => {
      panel[method](camRangeDiff);
    });

    return this;

}

  public movePanelElementsToStart(panels: ElementPanel[]) {
    const flicking = getFlickingAttached(this._flicking, "Renderer");
    const camera = flicking.camera;
    const cameraElement = camera.element;

    const panelEls = panels.map(panel => panel.element);
    const refElement = includes(panelEls, cameraElement.firstElementChild)
      ? null
      : cameraElement.firstElementChild;

    return this.movePanelElements(panels, refElement,"decreaseOffset");

  }

  public movePanelElementsToEnd(panels: ElementPanel[]) {
    return this.movePanelElements(panels,null,"increaseOffset");
  }
```

packages/vue-flicking/src/Flicking.vue
```diff
@@ -45,64 +68,74 @@ class VueFlicking extends Vue {
...
     const flicking = this._nativeFlicking;
-    const renderer = flicking.renderer as unknown as ExternalRenderer;
+    const renderer = flicking.renderer;
+
+    const slots = this._getSlots();
+    const diffResult = this._diffResult;
 
     diffResult.removed.forEach(idx => {
       renderer.remove(idx, 1);
+      console.log("removed", idx);
```
mixed: test?

src/core/Panel.ts
```diff
@@ -63,8 +68,34 @@ abstract class Panel {
     this._resetInternalStates();
   }
 
-  public abstract resize(): this;
-  public abstract contains(element: HTMLElement): boolean;
+  public resize(): this {
+    const el = this._el;
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+    const elStyle = window.getComputedStyle(el) || (el as any).currentStyle as CSSStyleDeclaration;
```
mixed, line 70: This code is often used. How about making it a method?

mixed (commented):
LGTM

src/renderer/Renderer.ts
WoodNeck: Well looks like you've selected only one commit in this PR.
That code is outdated and we're using `children` in the final commit.

packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
```diff
@@ -2,286 +2,233 @@
...
+          .map(panel => this.data[panel.index])
       });
-    } else {
-      this.renderPanelChange.emit(renderChangeEvent);
-    }
+    })
   }
 }
+
+export interface NgxFlickingComponent extends NativeFlicking {};
```
daybrush, line 234: There is an issue where angular or some compilers do not recognize the class.

So I created FlickingInterface.ts and inherited it.

src/Flicking.ts
```diff
@@ -35,14 +37,12 @@ export type FlickingEvents = {
   [EVENTS.NEED_PANEL]: NeedPanelEvent;
   [EVENTS.VISIBLE_CHANGE]: VisibleChangeEvent;
   [EVENTS.REACH_EDGE]: ReachEdgeEvent;
-  [EVENTS.CONTENT_ERROR]: void;
-};
+}
 
 /**
  * @param element A base element for the eg.Flicking module. When specifying a value as a `string` type, you must specify a css selector string to select the element.<ko>eg.Flicking 모듈을 사용할 기준 요소. `string`타입으로 값 지정시 요소를 선택하기 위한 css 선택자 문자열을 지정해야 한다.</ko>
```
daybrush, line 43: Does the interface look right in the doc documentation?

`@param` is used in methods, constructors, and events.
In the case of interface, `@typedef` is required and `@property` is used.

src/Flicking.ts
```diff
@@ -35,14 +37,12 @@ export type FlickingEvents = {
...
  * @param options An option object of the eg.Flicking module<ko>eg.Flicking 모듈의 옵션 객체</ko>
- * @param {string} [options.classPrefix="eg-flick"] A prefix of class names will be added for the panels, viewport, and camera.<ko>패널들과 뷰포트, 카메라에 추가될 클래스 이름의 접두사.</ko>
- * @param {number} [options.deceleration=0.0075] Deceleration value for panel movement animation for animation triggered by manual user input. A higher value means a shorter running time.<ko>사용자의 동작으로 가속도가 적용된 패널 이동 애니메이션의 감속도. 값이 높을수록 애니메이션 실행 시간이 짧아진다.</ko>
+ * @param {$ts:FlickingOptions["deceleration"]} [options.deceleration=0.0075] Deceleration value for panel movement animation for animation triggered by manual user input. A higher value means a shorter running time.<ko>사용자의 동작으로 가속도가 적용된 패널 이동 애니메이션의 감속도. 값이 높을수록 애니메이션 실행 시간이 짧아진다.</ko>
```
daybrush, line 45: Does the interface look right in the doc documentation?

`@param` is used in methods, constructors, and events.
In the case of interface, `@typedef` is required and `@property` is used.

packages/vue-flicking/src/Flicking.vue
```diff
@@ -0,0 +1,224 @@
...
+  public beforeUpdate() {
+    this._fillKeys();
+
+    const slots = this._getSlots();
+    const diffResult = this._slotDiffer.update(slots);
+    const flicking = this._nativeFlicking!;
+
+    this._diffResult = diffResult;
+
+    const removedPanels = diffResult.removed.reduce((map, idx) => {
```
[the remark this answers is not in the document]
WoodNeck: If the data nodes like `jsx` in React and `VNode` in Vue is not removed, removed should be an empty array.

src/renderer/RawRenderer.ts
WoodNeck: I'm not rushing to do it because there were a few changes in that code which makes it different from other parts.
I think merging them only makes it hard to change for future updates.
As it's a development phase for the v4, I think we can merge it later after the codes are stable.

src/renderer/RawRenderer.ts
WoodNeck: Well, codes are changed a bit from that point...

packages/vue-flicking/src/Flicking.vue
WoodNeck: Yep, I've removed that in the final commit.

src/core/Panel.ts
WoodNeck: Well it's the only part that uses that for calculating margin, so I can see any duplicates for it...

packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
WoodNeck: Is that an issue for the library users? I didn't have any issues while I'm building that.

daybrush: Yes. This will be a problem in angular's ivy compiler or product build.

src/Flicking.ts
WoodNeck: Oh, I just changed that because I've wanted to memo that `$ts:` in my code :)
I didn't fix & test jsdoc from v3, so you can ignore that.

src/renderer/RawRenderer.ts
mixed: If so, I would like to leave a record somewhere.

src/renderer/Renderer.ts
```diff
@@ -226,10 +226,17 @@ abstract class Renderer {
...
 
     const align = this._getPanelAlign();
 
+    // Remove all text nodes
+    cameraChilds.forEach(child => {
+      if (child.nodeType === Node.TEXT_NODE) {
```
daybrush: Perhaps it is text, comment, element that can be childNodes.

Since childNodes become panels, I think it's good to remove everything except elements.

WoodNeck: That's a good point, but does the comment node has a size(width) too?

daybrush: It seems that the comment node has no width.

Will the code that deletes the text node run in angular?

If so, it seems to be a problem.

WoodNeck: Well, actually you're seeing an outdated version of the Renderer.
The newer version uses `ElementManipulator` to do that, which isn't included in the Angular.

daybrush: Good. I think there's no problem if you use `children`.

daybrush (approved):
LGTM

src/renderer/RawRenderer.ts
WoodNeck: Sure, I'll leave a comment.

mixed (approved):

Issue #433, opened 2021-03-17
Flicking can move while the page's scrolling


WoodNeck:
## Description
Flicking can move while the page is scrolling.

## Steps to check or reproduce
- Click the Flicking, and scroll the page.
- While the page's scrolling move horizontally to see whether the Flicking's moving.

PR #434, opened 2021-03-17
fix: release immediately when the page's scrolling

### LINKED ISSUE #432: Carousel fires select event on touch devices even if scrolling up/down
## Description
Since the carousel is only detecting if the user is sliding left/right in the carousel, the user can accidentally trigger the 'select event' even if they are scrolling up and down on a touch device, accidentally dragging over the carousel.

1. Go to https://naver.github.io/egjs-flicking/features/events.html on a mobile device. (Flicking example for events)
2. Portrait mode, move carousel up and down, see events. (Move left to right, select will fire. Move up and down and not left to right, select will not fire)
3. Landscape mode, move carousel up and down, see events. (Move left to right, select will fire. Move up and down and not left to right, the page will move and select will fire.)

Notes:
The firing of the select event is both consistent and inconsistent, because the carousel itself is not sliding up and down, but in most cases on mobile, the select event can fire off when using your mobile device, and can cause problems. We use flicking to open a modal depending on the panel selected. While scrolling down on mobile, if the user tries scrolling past the flicking carousel, and drags their thumb over the carousel, the carousel shouldn't fire 'select', but does because it was not able to detect if the user was scrolling up/down. It just knows that the user did click on it, and the carousel didn't move left/right, even when the page moved up/down.

WoodNeck:
## Issue
#432, #433 

## Details
This fixes #432, #433

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "@egjs/flicking",
-	"version": "3.7.2-snapshot",
+	"version": "3.8.0-snapshot",
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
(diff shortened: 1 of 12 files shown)


mixed (approved):
LGTM

daybrush (approved):
LGTM

PR #435, opened 2021-03-19
test: add CFC e2e tests for v4

WoodNeck:
## Details
This adds some e2e tests built with codeceptjs + playwright and storybook


test/e2e/template/NgxFlickingTemplate.ts
```diff
@@ -0,0 +1,24 @@
+import { FlickingOptions } from "../../../src";
+import { NgxFlickingComponent } from "../../../packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component";
```
daybrush, line 2: If testing locally, it seems to be important whether to e2e after deploying native or before deploying native.


If you test before deployment, you may need a way to temporarily change the local version of the framework.

WoodNeck: That's right. That's why I've talked about dependency injection before.

daybrush: Maybe it will be resolved with resolve
https://webpack.js.org/configuration/resolve/

If not, there seems to be a way to install it from a local address.

mixed (approved):
LGTM

daybrush (approved):
LGTM

PR #441, opened 2021-04-16
fix: no current panel on the frameworks

### LINKED ISSUE #439: Auto height issue
## Description
How to make the slider responsive with auto height?

I have a dynamic content inside the slider. There might be slides with different height. if the content is very high I don't see part of the content. My slider is in the Dialog. Code looks like:

```
const flicking = React.useRef(null);

React.useEffect(() => {
    if (flicking.current) {
      flicking.current.resize();  //Seems it doesn't help also
    }
}, [props.slots]);

return (
  <Flicking
        style={{height: "100%"}} //without that style I don't see content at all
        gap={20}
        bound
        ref={flicking}
        lastIndex={60}
        infinite
        infiniteThreshold={400}
        onNeedPanel={handleNeedPanel}
        resizeOnContentsReady
      >
          {props.slots.map(slot => <div style={{width: 'calc(33% - 13px)'}}>some high content</div>)}
  </Flicking>
)
```

As you can see on the screenshot, there is not any scrollbar and part of the content is just hidden

27be7da144

is there any ideas how it could be improved?

WoodNeck:
## Issue
#439

## Details
This fixes #439

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "@egjs/flicking",
-	"version": "3.8.1",
+	"version": "3.8.1-snapshot",
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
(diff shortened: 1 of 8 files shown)


mixed (approved):
LGTM

daybrush (approved):
LGTM

PR #443, opened 2021-05-10
V4 docs

WoodNeck:
## Issue
#409


src/control/AxesController.ts
```diff
@@ -64,8 +117,15 @@ class AxesController {
...
       });
     }
+
+    return this;
   }
 
+  /**
+   * Destroy AxesController and return to initial state
+   * @ko AxesController를 초기 상태로 되돌립니다
+   * @return {void}
```
daybrush, line 127: @WoodNeck 

it's `return {this}`

daybrush (approved):
LGTM

mixed (approved):
👏👏👏

src/control/AxesController.ts
WoodNeck: Thanks~ I'll fix that

PR #447, opened 2021-05-27
fix: resize throws error when there're empty panels

WoodNeck:
## Details
This fixes issue that resize throws an error when there're empty panels between solid panels.


daybrush (approved):
LGTM

mixed (approved):
LGTM

PR #452, opened 2021-06-14
docs: add vue3-flicking to frameworks banner

WoodNeck:
## Details
This adds vue3-flicking banner to docs page

image


docs/src/pages/main/Frameworks.tsx
```diff
@@ -18,6 +20,10 @@ export default () => {
       <div className="framework-logo-wrapper mr-2"><img src="img/icons/vue.svg" /></div>
       <span>@egjs/vue-flicking</span>
     </div>
+    <div className={`framework-logo button mr-2 ${styles["is-vue3"]}`}>
```
mixed, line 23: image
image

I think that You have to add these two areas to the README for more people to see. :)

mixed (approved):
LGTM

docs/src/pages/main/Frameworks.tsx
WoodNeck: I'd like to, but isn't placing the same Vue icon twice looks bad on that page?

daybrush (approved):
LGTM

docs/src/pages/main/Frameworks.tsx
mixed: Um... I agree enough too. I think it's okay to just put a number like this, but what do you think?

vue2
vue3

WoodNeck: I think that's too much.
We should leave more space for later frameworks like Svelte or so

mixed: OK. Can you tell me what the appropriate guideline are for you?

WoodNeck: I mean, that one Vue logo can both mean Vue2 & Vue3

We can rather show a message like "if you're using Vu3, check vue3-flicking" in the vue-flicking README

mixed: That's a good way. You can link vue2 readme and vue3 readme. If you see the trend later and there are more vue3s, you can change the main link to vue3 and I think it's a good idea.

WoodNeck: Good, I've updated the README

PR #455, opened 2021-06-15
docs: add mr for github button

xch1029:

### DIFF
--- a/docs/src/pages/Main.tsx
+++ b/docs/src/pages/Main.tsx
@@ -23,7 +23,7 @@ export default () => <>
             🚀 Get Started
         </Link>
         <Link
-          className="button pl-5"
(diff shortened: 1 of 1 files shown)


WoodNeck:
Hello @xch1029!
That buttons are actually intentionally designed to stick together to show that the star is from Github, does that looks bad?

xch1029:
> Hello @xch1029!
> That buttons are actually intentionally designed to stick together to show that the star is from Github, does that looks bad?

now, I know the idea, It looks good. it just confused me the first time. now, I think it's a  intentionally  design. thank y for letting me know the design idea.

WoodNeck:
Thanks for the suggestion anyway, maybe I should remove the border-radius between those buttons :)

PR #456, opened 2021-06-15
docs: add source code to demos

WoodNeck:
## Details
This adds source codes to the demos page.

> light
image

> dark
image

### DIFF
--- a/docs/src/demo/Sync.tsx
+++ b/docs/src/demo/Sync.tsx
@@ -17,10 +17,10 @@ export default () => {
     flicking.camera.lookAt(flicking.camera.range.min + flicking.camera.rangeDiff * progress);
   };
 
-  const reactSourceCode = `export default (): React.FC => {
-  const flicking0 = useRef<Flicking>();
-  const flicking1 = useRef<Flicking>();
-  const flicking2 = useRef<Flicking>();
+  const reactSourceCode = `export default () => {
+  const flicking0 = useRef();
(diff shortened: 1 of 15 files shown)


mixed (approved):
AWESOME!

daybrush:
@WoodNeck 

Vanilla has flicking import code.

There is no import part in the rest of the framework.

it's ok?

WoodNeck:
@daybrush 
I think I put that because I wanted to show importing constants.
It's optional and trivial, I think

docs/src/demo/Sync.tsx
```diff
@@ -13,7 +17,171 @@ export default () => {
     flicking.camera.lookAt(flicking.camera.range.min + flicking.camera.rangeDiff * progress);
   };
 
-  return <div className="has-background-grey-lighter p-4">
+  const reactSourceCode = `export default (): React.FC => {
+  const flicking0 = useRef<Flicking>();
```
daybrush, line 21: In the past, there was a question that a javascript user copied and pasted the typescript code and it didn't work.

Since angular is forced typescript, there is no problem, but in react, how about omitting the type for javascript users?

daybrush (approved):
LGTM

docs/src/demo/Sync.tsx
WoodNeck: Oh, I didn't think that.
I'll fix that to use js instead then

PR #478, opened 2021-07-02
fix: preventClickOnDrag to work on non-native clickable elements

### LINKED ISSUE #470: purpose of preventClickOnDrag ?
## Description
Can someone explain a bit more what [preventClickOnDrag](https://naver.github.io/egjs-flicking/docs/api/Flicking/#preventClickOnDrag) does? A click event for me is indeed bubbling up the DOM whenever I drag the carousel -- with `preventClickOnDrag` set to true, is the click event not suppose to bubble up?

These are my dependencies:
```
 "@egjs/flicking": "^4.1.0",
 "@egjs/vue-flicking": "^4.1.0"
 ```
 
 I'm currently using Vue 2.6.11 and typescript 3.7.4

### PR DESCRIPTION
**Pull request by WoodNeck**
## Issue
#470

## Details
This fixes #470

### DIFF
--- a/src/control/AxesController.ts
+++ b/src/control/AxesController.ts
@@ -24,6 +24,7 @@ class AxesController {
   private _stateMachine: StateMachine;
 
   private _animatingContext: { start: number; end: number; offset: number };
+  private _dragged: boolean;
 
   /**
    * An {@link https://naver.github.io/egjs-axes/release/latest/doc/eg.Axes.html Axes} instance
@@ -128,6 +129,8 @@ class AxesController {
    * @return {void}
    */
   public destroy(): void {
+    this.removePreventClickHandler();
+
     this._axes?.destroy();
     this._panInput?.destroy();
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/control/AxesController.ts [hunk 1]
```diff
@@ -183,6 +186,26 @@ class AxesController {
...
   }
 
+  public addPreventClickHandler() {
+    const flicking = getFlickingAttached(this._flicking, "Control");
+    const axes = this._axes!;
+    const viewportEl = flicking.element;
+
+    axes.on(AXES.EVENT.HOLD, this._onAxesHold);
+    axes.on(AXES.EVENT.CHANGE, this._onAxesChange);
+    viewportEl.addEventListener("click", this._preventClickWhenDragged, true);
```
daybrush, line 196: Do you know how far to prevent the click event?

If a click event is prevented even if the drag pointer leaves the viewport, it seems to be necessary to add a click capturing event targeting the window.

daybrush (approved):
LGTM

mixed (approved):
LGTM

src/control/AxesController.ts
WoodNeck: In what specific case do you mean?
Point of the click event preventing is for panel elements, so I think it's okay as they're all in the viewport elements.

PR #489, opened 2021-07-14
feat: add hideBeforeInit & firstPanelSize

### LINKED ISSUE #457: Align and renderOnlyVisible with SSR
## Description
Flicking v4 is positioned as SSR-ready library. Trying to make it work with SSR i got some questions.

My setup is like 
```tsx
<Flicking
    align={{ panel: '50%', camera: 0 }}
    circular={true}
    renderOnlyVisible={true}
    ...
```

1) Question about align prop. With my setup i have first seen panel is half off the screen.  It seems that this 50% shift is done by `transform: translate` css property which can not be calculated on SSR, so initial markup does not have this shift, but after hydration it gets shifted. Does it mean i can use only "prev" align value on SSR to match markup?

2) Question about renderOnlyVisible prop. I can have 100+ possible panels so i want to render only visible of them. Initial markup from SSR contains every panel, after hydration only visible panels remain. As number of panels is large, i don't want to send markup of all 100+ panels to client. Is it possible to create only **n** panels on SSR like 10? i can slice items array by myself on server and after hydration use full array, but maybe it is possible by flicking?

WoodNeck:
## Issue
#457 

## Details
This resolves #457


docs/docs/ssr.mdx
```diff
@@ -0,0 +1,107 @@
+---
```
mixed, line 1: GOOD DOCS.

mixed (approved):
LGTM

daybrush (approved):
LGTM

PR #490, opened 2021-07-14
feat: add noEventsBeforeInit

### LINKED ISSUE #486: v4 - Triggers `onMoveEnd` event at initial rendering
## Description

`onMoveEnd` event is triggered at initial rendering on v4. v3 does not trigger `onMoveEnd` at initial rendering. In my opinion, user did not move anything, so event should not be triggered. Did I miss some configuration? or, is this behavior intended?

See the console log in [this codesandbox](https://codesandbox.io/s/egjsflicking-v4-on-move-end-43kmk?file=/src/App.js)

WoodNeck:
## Issue
#486 

## Details
This resolves #486

### DIFF
--- a/src/Flicking.ts
+++ b/src/Flicking.ts
@@ -53,6 +53,7 @@ export interface FlickingOptions {
   adaptive: boolean;
   // EVENT
   needPanelThreshold: number;
+  noEventsBeforeInit: boolean;
   // ANIMATION
   deceleration: number;
   duration: number;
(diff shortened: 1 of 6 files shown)


mixed (approved):
LGTM

src/Flicking.ts
```diff
@@ -53,6 +53,7 @@ export interface FlickingOptions {
   adaptive: boolean;
   // EVENT
   needPanelThreshold: number;
+  noEventsBeforeInit: boolean;
```
mixed, line 56: Longer name than `noEventsBeforeInit`, but what about `preventEventsBeforeInit`?

WoodNeck: I chose that first time, but I've changed like that as you guys think it'd too long and thought it could be confusing with native `preventDefault`
@daybrush How do you think?

daybrush: If the name itself is not the same, there seems to be no problem. `preventEventsBeforeInit` is good.

daybrush (approved):
LGTM

src/Flicking.ts
WoodNeck: Well okay then, I'll change it like that 👍

PR #499, opened 2021-07-27
fix: viewport size calculation

### LINKED ISSUE #498: Flicking's panel placement can be broken when there's border in the viewport element
## Description
As Flicking uses `offsetWidth` instead of `clientWidth` for calculating viewport element's size, panel placement can be broken when there's CSS border applied on the viewport element.

WoodNeck:
## Issue
#498 

## Details
This fixes #498


daybrush (approved):
LGTM

PR #502, opened 2021-07-28
fix: set axes releaseOnScroll to true

WoodNeck:
## Details
This restores missing `releaseOnScroll` option for Axes that was previously used in Flicking v3


mixed (approved):
LGTM

daybrush (approved):
LGTM

PR #509, opened 2021-08-06
fix: set new currentPanel when current is removed

WoodNeck:
## Details
This fixes case that flicking doesn't update `currentPanel` after calling `sync`


mixed (approved):
LGTM

daybrush (approved):
LGTM

PR #510, opened 2021-08-09
fix: cfc sync's removing algorithm

### LINKED ISSUE #506: Dynamic children errors with `Cannot read property 'key' of undefined`
## Description
Having a dynamic children will cause errors in flicking. Is this meant to be supported? As a workaround, you can generate a key based on the children, so it will mount a new component when children change, but this is not ideal.

In this sandbox, the children are randomly generated every second. If you remove the `key` on `<Flicking>`, it will cause an error pretty quickly.
https://codesandbox.io/s/restless-butterfly-ttl8g?file=/src/App.tsx

WoodNeck:
## Issue
#506 

## Details
This fixes #506

### DIFF
--- a/src/cfc/sync.ts
+++ b/src/cfc/sync.ts
@@ -10,23 +10,23 @@ export default (flicking: Flicking, diffResult: DiffResult<any>, rendered: any[]
     let startIdx = -1;
     let prevIdx = -1;
 
-    diffResult.removed.forEach((removedIdx, idx) => {
+    [...diffResult.removed].reverse().forEach(removedIdx => {
       if (startIdx < 0) {
-        startIdx = idx;
+        startIdx = removedIdx;
       }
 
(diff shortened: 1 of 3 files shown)


src/cfc/sync.ts
```diff
@@ -10,23 +10,23 @@ export default (flicking: Flicking, diffResult: DiffResult<any>, rendered: any[]
     let startIdx = -1;
     let prevIdx = -1;
 
-    diffResult.removed.forEach((removedIdx, idx) => {
+    [...diffResult.removed].reverse().forEach(removedIdx => {
```
daybrush, line 13: If the order is correct, the index will change every time you delete it, is that ok?

daybrush (approved):
LGTM

src/cfc/sync.ts
WoodNeck: Good point, I should add a TC for that

mixed (approved):
LGTM

PR #515, opened 2021-08-17
fix: bound not working when there're not enough panels to fill the viewport

### LINKED ISSUE #505: bound=true + align="prev" with few children not working correctly
## Description
When you have a small number of children and the settings `bound={true}` and `align="prev"`, the carousel allows advancing beyond the bound edge. When the number of children is less than a full viewport, it should not advance at all.

Example:
https://codesandbox.io/s/nice-cannon-hhio5?file=/src/App.tsx

WoodNeck:
## Issue
#505 

## Details
This fixes #505 

"prev"
image

"next"
image

"center"
image

"20%"
image

### DIFF
--- a/src/camera/BoundCamera.ts
+++ b/src/camera/BoundCamera.ts
@@ -4,7 +4,7 @@
  */
 import Panel from "../core/panel/Panel";
 import AnchorPoint from "../core/AnchorPoint";
-import { getFlickingAttached } from "../utils";
+import { getFlickingAttached, parseAlign } from "../utils";
 
 import Camera from "./Camera";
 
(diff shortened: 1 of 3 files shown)


mixed (approved):
LGTM

src/camera/BoundCamera.ts
```diff
@@ -39,12 +39,22 @@ class BoundCamera extends Camera {
...
+    if (isBiggerThanViewport) {
+      this._range = { min: firstPos, max: lastPos };
     } else {
-      this._range = { min: firstPanel.position, max: lastPanel.position };
+      const align = this._align;
+      const alignVal = typeof align === "object"
+        ? (align as { camera: string | number }).camera
+        : align;
+
+      const pos = firstPos + parseAlign(alignVal, lastPos - firstPos);
```
daybrush, line 55: If there are two panels with a width of 30% and the align is `{ camera: "50%", panel: 0 }`, is it placed in a normal position?

It looks like the panel will go out of the viewport area.

daybrush (approved):
LGTM

src/camera/BoundCamera.ts
WoodNeck: image
I've test that, and it looks like the above.

PR #521, opened 2021-08-19
fix: setPrototypeOf is not defined on IE10 or lower

### LINKED ISSUE #518: `setPrototypeOf` is not defined on IE10 or lower
https://developer.mozilla.org/ko/docs/Web/JavaScript/Reference/Global_Objects/Object/setPrototypeOf
Should be polyfilled

WoodNeck:
## Issue
#518 

## Details
This fixes #518

### DIFF
--- a/src/core/FlickingError.ts
+++ b/src/core/FlickingError.ts
@@ -3,6 +3,8 @@
  * egjs projects are licensed under the MIT license
  */
 
+import { setPrototypeOf } from "../utils";
+
 /**
  * Special type of known error that {@link Flicking} throws.
  * @ko Flicking 내부에서 알려진 오류 발생시 throw되는 에러
@@ -30,7 +32,8 @@ class FlickingError extends Error {
     public message: string,
     public code: number) {
     super(message);
-    Object.setPrototypeOf(this, FlickingError.prototype);
+
+    setPrototypeOf(this, FlickingError.prototype);
     this.name = "FlickingError";
   }
 }
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -265,3 +265,8 @@ export const circulateIndex = (index: number, max: number): number => {
     return index;
   }
 };
+
+export const setPrototypeOf = Object.setPrototypeOf || ((obj, proto) => {
+  obj.__proto__ = proto;
+  return obj;
+});
--- a/test/manual/js/index.js
+++ b/test/manual/js/index.js
@@ -1,4 +1,4 @@
-const flicking = new Flicking("#flicking", {
+var flicking = new Flicking("#flicking", {
   align: "10%",
   bound: true
 });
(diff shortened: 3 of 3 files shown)


src/core/FlickingError.ts
```diff
@@ -30,7 +32,8 @@ class FlickingError extends Error {
     public message: string,
     public code: number) {
     super(message);
-    Object.setPrototypeOf(this, FlickingError.prototype);
+
+    setPrototypeOf(this, FlickingError.prototype);
```
daybrush, line 36: By doing setPrototypeOf, what functionality do you expect?

daybrush (approved):
LGTM

mixed (approved):
LGTM

src/core/FlickingError.ts
WoodNeck: While making this, there were a few restrictions. I don't remember everything but here's some
- a thrown `err instanceof FlickingError` should be true
- a thrown `err instanceof Error` also should be true
- FlickingError should have `message`, `stack` and `code` property
- FlickingError should be throwable & catchable

daybrush: Can it be seen that this cannot be solved by simply inheriting the Error class?

WoodNeck: Yes, that doesn't work for some reason...
I remember it couldn't solve one of the restrictions above, I guess the first one was?

daybrush: I don't know why. Is it because it is not a native class because it is transpiled?

WoodNeck: If I remove that, `expect(err).to.be.instanceOf(FlickingError);` fails.

As you said, it's not a native class after it's transpiled.

> BEF remove
```js
    var FlickingError = function (_super) {
      __extends$1(FlickingError, _super);
      /**
       * @param message Error message<ko>에러 메시지</ko>
       * @param code Error code<ko>에러 코드</ko>
       */


      function FlickingError(message, code) {
        var _this = _super.call(this, message) || this;

        _this.message = message;
        _this.code = code;
        setPrototypeOf(_this, FlickingError.prototype);
        _this.name = "FlickingError";
        return _this;
      }

      return FlickingError;
    }(Error);
```

> AFT remove
```js
    var FlickingError = function (_super) {
      __extends$1(FlickingError, _super);
      /**
       * @param message Error message<ko>에러 메시지</ko>
       * @param code Error code<ko>에러 코드</ko>
       */


      function FlickingError(message, code) {
        var _this = _super.call(this, message) || this;

        _this.message = message;
        _this.code = code;
        _this.name = "FlickingError";
        return _this;
      }

      return FlickingError;
    }(Error);
```


```js
    function __extends$1(d, b) {
      if (typeof b !== "function" && b !== null) throw new TypeError("Class extends value " + String(b) + " is not a constructor or null");
      extendStatics$1(d, b);

      function __() {
        this.constructor = d;
      }

      d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
    }
```

By that, `expect(err).to.be.instanceOf(FlickingError);` fails.

PR #523, opened 2021-08-23
fix: needPanel triggers multiple times when panel is appended by event callback

### LINKED ISSUE #522: needPanel triggers multiple times when panel is appended by event callback
## Description
`needPanel` triggers multiple times when panel is appended by event callback

```js
flicking.on(Flicking.EVENTS.NEED_PANEL, e => {
  if (e.direction === "PREV") return;

  flicking.append(`<div class="flicking-panel has-text-white has-background-primary">${flicking.panelCount + 1}</div>`);
});

```

WoodNeck:
## Issue
#522 

## Details
This fixes #522

### DIFF
--- a/src/control/AxesController.ts
+++ b/src/control/AxesController.ts
@@ -36,6 +36,10 @@ class AxesController {
    */
   public get axes() { return this._axes; }
   /**
+   * @internal
+   */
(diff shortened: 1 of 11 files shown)


daybrush (approved):
LGTM

PR #532, opened 2021-08-25
fix: angular peer dependency not updated

### LINKED ISSUE #530: incompatible with angular 12.2.2
## Description
When I try to update my angular project to 12.2.2 by using `ng update` command, I get:
`
Package "@egjs/ngx-flicking" has an incompatible peer dependency to "@angular/common" (requires "^7.2.0" (extended), 
would install "12.2.2").
Package "@egjs/ngx-flicking" has an incompatible peer dependency to "@angular/core" (requires "^7.2.15" (extended), would 
install "12.2.2").

× Migration failed: Incompatible peer dependencies found.
Peer dependency warnings when installing dependencies means that those dependencies might not work correctly together.
You can use the '--force' option to ignore incompatible peer dependencies and instead address these warnings later.

`

WoodNeck:
## Issue
#530

## Details
This fixes #530


mixed (approved):
LGTM

packages/ngx-flicking/projects/ngx-flicking/package.json
```diff
@@ -3,8 +3,8 @@
   "version": "4.2.3",
   "description": "Everyday 30 million people experience. It's reliable, flexible and extendable carousel.",
   "peerDependencies": {
-    "@angular/common": "^7.2.0",
-    "@angular/core": "^7.2.15"
+    "@angular/common": ">=11.0.0",
```
daybrush, line 6: It seems that all versions 7, 8, 9, 10, and 11 should be included.

WoodNeck: I've excluded 7~9 as it's not supported as LTS in Angular community.
I wish to it be 10+, but the current version works only in the 11+, maybe I should recreate the whole project to support that.

Also, I'm wondering that if the library supports Angular major 7, does that also mean the other versions(8+)?
I'm worried that it might not support those latest versions if I built it with Angular 7

WoodNeck:
https://semver.npmjs.com/

PR #537, opened 2021-08-27
fix: vue3 render triggers update

### LINKED ISSUE #535: Vue3 recursion limit exceeded
## Description
I'm using the flicking component to render a list from the demo examples but I keep getting a `Maximum recursive updates exceeded. This means you have a reactive effect that is mutating its own dependencies and thus recursively triggering itself. Possible sources include component template, render function, updated hook or watcher source function.`  warning. 

This breaks the built version of my application and causes the browser tab to freeze. 

```html
<Flicking :options="{ circular: true }">
  <div v-for="idx in list" class="flicking-panel w-80" :key="idx">{{ idx }}</div>
</Flicking>
```

```javascript
import Flicking from "@egjs/vue3-flicking";
import "@egjs/vue3-flicking/dist/flicking.css";

components: {
     Flicking
},
data() {
  return {
    list: [0, 1, 2, 3, 4],
  }
}
```

Not sure if it's an issue with my implementation or something else.

WoodNeck:
## Issue
#535

## Details
This fixes #535

### DIFF
--- a/packages/vue3-flicking/package-lock.json
+++ b/packages/vue3-flicking/package-lock.json
@@ -65,7 +65,8 @@
     "@babel/helper-validator-identifier": {
       "version": "7.14.5",
       "resolved": "https://registry.npmjs.org/@babel/helper-validator-identifier/-/helper-validator-identifier-7.14.5.tgz",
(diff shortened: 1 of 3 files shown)


mixed (approved):
LGTM

PR #542, opened 2021-09-01
fix: Flicking is animating on moveEnd event

### LINKED ISSUE #541: Flicking is animating on moveEnd event
## Description
- https://github.com/naver/egjs-flicking-plugins/issues/26

WoodNeck:
## Issue
#541

## Details
This fixes #541

### DIFF
--- a/src/camera/Camera.ts
+++ b/src/camera/Camera.ts
@@ -271,11 +271,11 @@ abstract class Camera {
    * <ko>{@link ERROR_CODE NOT_ATTACHED_TO_FLICKING} {@link Camera#init init}이 이전에 호출되지 않은 경우</ko>
    * @return {this}
    */
-  public async lookAt(pos: number): Promise<void> {
+  public lookAt(pos: number): void {
     const prevPos = this._position;
 
     this._position = pos;
-    await this._refreshVisiblePanels();
+    this._refreshVisiblePanels();
     this._checkNeedPanel();
     this._checkReachEnd(prevPos, pos);
     this._applyTransform();
@@ -514,7 +514,7 @@ abstract class Camera {
     this._needPanelTriggered = { prev: false, next: false };
   }
 
-  protected async _refreshVisiblePanels() {
+  protected _refreshVisiblePanels() {
     const flicking = getFlickingAttached(this._flicking, "Camera");
     const panels = flicking.renderer.panels;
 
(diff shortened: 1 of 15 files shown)


src/camera/Camera.ts
```diff
@@ -271,11 +271,11 @@ abstract class Camera {
    * <ko>{@link ERROR_CODE NOT_ATTACHED_TO_FLICKING} {@link Camera#init init}이 이전에 호출되지 않은 경우</ko>
    * @return {this}
    */
-  public async lookAt(pos: number): Promise<void> {
+  public lookAt(pos: number): void {
     const prevPos = this._position;
 
     this._position = pos;
-    await this._refreshVisiblePanels();
+    this._refreshVisiblePanels();
```
mixed, line 278: The logic seems to be different from the previous one. I wonder if that was part of the intention. The point where EVENTS.VISIBLE_CHANGE occurs after flicking.renderer.render() is finished is the same as before. However, the three methods below can be executed before VISIBLE_CHANGE occurs differently from the existing ones. I wonder if there are any problems with this part.
```js
this._checkNeedPanel();
this._checkReachEnd(prevPos, pos);
this._applyTransform();
```

mixed (commented):
If there are no problems with the `lookat` review, this is a good enough PR.

src/camera/Camera.ts
```diff
@@ -526,13 +526,13 @@ abstract class Camera {
...
 
     if (added.length > 0 || removed.length > 0) {
-      await flicking.renderer.render();
-
-      flicking.trigger(new ComponentEvent(EVENTS.VISIBLE_CHANGE, {
-        added,
-        removed,
-        visiblePanels: newVisiblePanels
-      }));
+      void flicking.renderer.render().then(() => {
```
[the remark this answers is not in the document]
WoodNeck: That's basically due to our eslint setting

src/camera/Camera.ts
WoodNeck: Yes, the timing changed and even the interface of the internal component is changed.
But as there wasn't any previous definition of the timing of `visibleChange`, I hope that's okay.

mixed (approved):
LGTM

daybrush (approved):
LGTM

PR #547, opened 2021-09-02
feat: add resizeOnContentsReady

### LINKED ISSUE #546: New feature: resizeOnContentsReady
## Description
Enabling this option will automatically call Flicking's resize when all image/video inside panels are loaded.
This can be useful when you have contents inside Flicking that changes its size when it's loaded

### LINKED ISSUE #548: currentPanel is null when the initial panel size is 0
## Description
This is caused as Axes doesn't trigger any events when the start position is same to the end position.

WoodNeck:
## Issue
#546 

## Details
This resolves #546 
This fixes #548


mixed (approved):
GOOD. Feature!

daybrush (approved):
LGTM

PR #549, opened 2021-09-03
docs: rollback lazyload demo

WoodNeck:
## Details
This will remove the pre-deployed lazyload demo, as it's not properly prepared yet.
That demo can have memory issues with old devices, so I'll remove that demo for now, and will add it again when it's prepared.


mixed (approved):
LGTM

daybrush (approved):
LGTM

PR #559, opened 2021-10-05
feat: add virtual

WoodNeck:
## Details
This adds a new option named virtual


mixed (approved):
LGTM.
It's a good feature, but it seems to have affected the overall code as virtual is implemented. After the work is completed, it would be good to think about whether the current method is appropriate.

PR #586, opened 2021-11-12
fix: vue3-flicking can't refresh panel element

### LINKED ISSUE #585: Fade plugin won't work with renderOnlyVisible [vue3-flicking]
## Description
Hi,
Fade plugin tries to modify the original panel style but the diffResult won't carry them so the `opacity` was never actually set for Vue elements.

Sorry for missing the repo. I'm in the middle of something. Will provide soon if it's not clear for you.

WoodNeck:
## Issue
#585

## Details
This fixes #585

### DIFF
--- a/packages/vue3-flicking/demo/App.vue
+++ b/packages/vue3-flicking/demo/App.vue
@@ -71,16 +71,12 @@
           <span class="flicking-arrow-next"></span>
         </template>
       </Flicking>
-      <Flicking :plugins="arrow">
+      <Flicking :plugins="fade" :options="{ renderOnlyVisible: true }">
         <div class="panel">0</div>
         <div class="panel">1</div>
         <div class="panel">2</div>
         <div class="panel">3</div>
         <div class="panel">4</div>
-        <template #viewport>
-          <span class="flicking-arrow-prev"></span>
-          <span class="flicking-arrow-next"></span>
-        </template>
       </Flicking>
       <Flicking class="flicking flicking0" ref="vFlick" :options="{
         panelsPerView: 5,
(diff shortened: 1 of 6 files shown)


packages/vue3-flicking/src/VueElementProvider.ts
```diff
@@ -8,9 +8,20 @@ import VuePanel from "./VuePanel";
...
 
-  public get element() { return this._el.nativeElement; }
   public get rendered() { return !this._el.hide; }
+  public get element() {
+    const el = this._el.$el.nextSibling as HTMLElement;
+
+    if (!el || el.nodeType !== Node.ELEMENT_NODE) {
+      return this._cachedElement;
+    } else {
+      this._cachedElement = el;
```
mixed, line 20: How about it?
```js
if (el && el.nodeType === Node.ELEMENT_NODE) {
    this._cachedElement = el;
}

return this._cachedElement;
```

mixed (approved):
LGTM

packages/vue3-flicking/src/VueElementProvider.ts
WoodNeck: @mixed 
That looks much better 👍

packages/vue3-flicking/demo/App.vue
```diff
@@ -71,16 +71,12 @@
           <span class="flicking-arrow-next"></span>
         </template>
       </Flicking>
-      <Flicking :plugins="arrow">
+      <Flicking :plugins="fade" :options="{ renderOnlyVisible: true }">
```
daybrush, line 74: Can't the demo use both fade and arrow?

daybrush (approved):
LGTM

packages/vue3-flicking/demo/App.vue
WoodNeck: It can, but I just separated it to test the combination of fade + renderOnlyVisible

PR #589, opened 2021-11-17
test: add CFC tests

### LINKED ISSUE #583: preact-flicking is not working
## Description
export '_unmount' was not found in 'preact'
export 'isFragment' was not found in 'preact'

Can be fixed with preact version update inside preact-flicking.

WoodNeck:
## Details
This adds CFC tests for Flicking frameworks
- Working Github actions: https://github.com/WoodNeck/egjs-flicking/runs/4235347692?check_suite_focus=true
- Coveralls: https://coveralls.io/github/WoodNeck/egjs-flicking

This fixes #583


mixed (approved):
LGTM

packages/svelte-flicking/src/flicking-panel.svelte
```diff
@@ -3,9 +3,9 @@
     onDestroy,
     getContext
   } from "svelte";
-  import * as uuid from "uuid-browser";
+  import { v4 } from "uuid";
```
daybrush, line 6: Why did you change from uuid-browser to uuid?

WoodNeck: I believe that was because `uuid-browser` couldn't run in test env.
While the `uuid` can run in both env.

daybrush: uuid contains binary modules. It will be difficult to work in the browser.

daybrush: image

test/cfc/framework/vue3/Vue3FixtureRenderer.tsx
```diff
@@ -0,0 +1,82 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+const { defineComponent, h, resolveComponent } = require("vue");
```
daybrush, line 2: Is there any reason why only vue doesn't use import?

packages/svelte-flicking/src/flicking-panel.svelte
WoodNeck: Well if you're using the esm module just with CDN like that, you're right.
But in svelte we're using it with bundlers by default, aren't we?
They even say they can use it in the browsers.
- https://www.npmjs.com/package/uuid#user-content-ecmascript-modules

Also, I've tested in both my local dev-env & other env that use standard svelte dev-env.

test/cfc/framework/vue3/Vue3FixtureRenderer.tsx
WoodNeck: Actually, that was needed at some moment that I can't remember.
But, looks like changing that doesn't trigger any test error. I'll change that to normal esm import.

daybrush (approved):
LGTM

ahn3407 (approved):
LGTM

PR #590, opened 2021-11-18
refactor: add externalRenderer

WoodNeck:
## Details
- This fixes internal typings by adding a new option to Flicking - `externalRenderer`
- This deprecates option `renderExternal`
- This fixes Vue & Vue3 Flicking's type

This is marked as `refactor` because both options are internal-only


mixed (approved):
LGTM

packages/vue3-flicking/src/Flicking.ts
```diff
@@ -75,23 +126,19 @@ class Flicking extends Vue.with(FlickingProps) {
     if (this.status) {
       flicking.setStatus(this.status);
     }
-  }
-
-  public beforeDestroy() {
+  },
+  beforeDestroy() {
```
daybrush, line 130: I think you can use beforeUnmount.

WoodNeck: Looks like I should, thanks for checking 👍

daybrush (approved):
LGTM

ahn3407 (approved):
LGTM

PR #592, opened 2021-11-18
fix: direction is always null after disabling input

### LINKED ISSUE #591: V3 - direction is null after disabling input
## Description
After calling `disableInput` while holding and re-enabling the input, the next event's property `direction` is always null.

1. Call `disableInput` while holding the input
2. Call `enableInput` (no matter called immediately or not)

WoodNeck:
## Issue
#591

## Details
This fixes #591

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -211,6 +211,7 @@ temp/
 demo/release
 packages/*/dist/
 demo/_data/version.yml
+lib/
 dist/
 outjs/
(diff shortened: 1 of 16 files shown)


mixed (approved):
LGTM

ahn3407 (approved):
LGTM

daybrush (approved):
LGTM

Issue #607, opened 2022-01-12
Circle and bound if  there are fewer slides than viewport


lna1989:
## Description
If circular:true and bound:true with the number of slides less than the width/height viewport bound: true option does not work (But in theory, the circular option turns from true to false in this case)

## Steps to check or reproduce
[codesandbox](https://codesandbox.io/s/flicking-circulart-and-bound-t83no?file=/src/App.vue)

WoodNeck:
Hello @lna1989!
Both are physically impossible, as they can't fill the viewport area if the sum of panel sizes is smaller than the viewport size.
So we're turning off such options to prevent graphical glitches.
The only solution is to clone the panel elements. We did that in the previous major version(v3), but we're not doing this automatically in newer versions for element consistency (like, to prevent events from the cloned elements)

Maybe the `bound` can be improved, like restricting the scrollable area to 0, but I think `circular` can't without graphical glitches so we have to leave it as is.
If you're aware of panel/viewport sizes, you can try cloning the panel elements.

lna1989:
Hello @WoodNeck! 
I mean the case when circle is actually turned off, even though the option is enabled. I.e. fliking determines that the viewport is not fully filled and turns off the circle() option:

> WARNING
If the sum of the panel sizes is too small, circular will not be enabled.
This is because Flicking doesn't clone panel elements for the performance and to prevent other issues.
You can check this with circularEnabled property of Flicking.
To enable circular mode, all panels should pass the below condition
Sum of panel sizes - panel size >= viewport size

In the same case, we can enable the binding: true , since the circle is actually disabled

WoodNeck:
That's interesting. You mean you want to enable `bound` for the case when `circular` can't be enabled?

lna1989:
> That's interesting. You mean you want to enable `bound` for the case when `circular` can't be enabled?

@WoodNeck Absolutely righ, attached a real case where it can be useful
![Peek 2022-01-13 09-51](https://user-images.githubusercontent.com/49455662/149273554-242648c1-17f2-4e04-a252-dac5ef880750.gif)
In theory, I can check the option on my side and send circle to false, and bound to true, but I think this is not the most interesting solution, since flicking already knows how to send circle to false.

WoodNeck:
Now I fully understood the situation, thanks for the detailed explanation.

I think I can add an option for that, named like `circularFallback` as using `bound` for fallback is not intuitive in this case I think.
[...]

PR #610, opened 2022-01-17
feat: add circularFallback

WoodNeck:
## Issue
#607 

## Details
This adds new option `circularFallback` and two available values for it: `default` & `bound`


src/camera/Camera.ts
```diff
@@ -637,7 +687,7 @@ abstract class Camera {
     }));
   }
 
-  protected _checkTranslateSupport = () => {
+  private _checkTranslateSupport = () => {
```
daybrush, line 690: (protected to private) It would be nice if it wasn't exposed to the outside.

src/const/external.ts
```diff
@@ -98,3 +98,15 @@ export const CLASS = {
...
+
+/**
+ * An object with all possible {@link Flicking#circularFallback circularFallback}s
+ * @ko Flicking의 {@link Flicking#circularFallback circularFallback}에 설정 가능한 값들을 담고 있는 객체
+ * @type {object}
+ * @property {string} DEFAULT "default"
+ * @property {string} BOUND "bound"
+ */
+export const CIRCULAR_FALLBACK = {
+  DEFAULT: "default",
```
daybrush, line 110: If circular does not work, it seems that i do not understand the meaning of `default`. How about `linear`? I confirmed that the camera name is LinearCamera.

daybrush (approved):
LGTM

malangfox (approved):
LGTM

src/const/external.ts
WoodNeck: I chose that name because I thought that `linear` is just an internal term for defining the default scrolling type.
But, I think linear is also okay.

src/camera/Camera.ts
WoodNeck: As it's `protected` it was open to its child classes(`LinearCamera` and so).
But as they're all gone now, it's bound to the `Camera` only.

PR #626, opened 2022-02-14
fix: can't initialize when all panels have size 0

### LINKED ISSUE #624: Fix on 3.9.1 made the whole page freeze (relative issue: #616)
"@egjs/vue-flicking": "3.8.1"
## Description
Hello team. Thanks a lot for the fix! 
However, this fix introduced a much bigger bug. When the scenario of the [#616](https://github.com/naver/egjs-flicking/issues/616) happen, then the whole page freezes due to some endless loop happening next.
After debugging it, we noticed that `panel0OnFirstScreen` equals to `NaN` due to `0/0 division` because all the variables on the right hand side equal to 0. This makes the `cloneCount` to equal to `NaN`, creating a domino effect afterwards with endless loops.
image

A suggested fix that seemed to resolve the domino effect is to add a fallback on `cloneCount`, but we are not sure if this is the correct one.
image

We were forced to downgrade to `"@egjs/vue-flicking": "3.8.0"` with `egjs-flicking: 3.9.0`, because the error that was thrown in the previous version saves as from the whole page to freeze.

WoodNeck:
## Issue
#624

## Details
This fixes #624

### DIFF
--- a/src/components/Viewport.ts
+++ b/src/components/Viewport.ts
@@ -1265,8 +1265,10 @@ export default class Viewport {
     // For each panels, clone itself while last panel's position + size is below viewport size
     const panels = panelManager.originalPanels();
     const reversedPanels = panels.concat().reverse();
-
     const sumOriginalPanelSize = lastPanel.getPosition() + lastPanel.getSize() - firstPanel.getPosition() + gap;
+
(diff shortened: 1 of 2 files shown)


malangfox (approved):
LGTM

PR #653, opened 2022-04-14
fix: wrong sync if ordered is included

WoodNeck:
## Details
This fixes the broken `sync` implementation when an entry of `ordered` is included in the diff result.


qpitlove (approved):

daybrush (approved):
LGTM

PR #656, opened 2022-04-19
fix: no camera pos update when animating

WoodNeck:
## Details
This fixes the framework error that the camera position resets to 0 when updating.


src/renderer/Renderer.ts
```diff
@@ -337,24 +337,26 @@ abstract class Renderer {
...
+          }
+
+          void control.moveToPanel(panels[targetIndex], {
+            duration: 0
+          }).catch(() => void 0);
         }
-
-        void control.moveToPanel(panels[targetIndex], {
+      } else {
+        void control.moveToPanel(control.activePanel!, {
```
daybrush, line 356: Is it like the activePanel constant? Or is the value of control.activePanel changing?

daybrush (approved):
LGTM

qpitlove (approved):

src/renderer/Renderer.ts
WoodNeck: I think that's just a mistake! I'll change it to `activePanel`

malangfox (approved):
LGTM

PR #670, opened 2022-05-10
feat: add nested option

malangfox:
## Details
nestedgrid

`nested` is added to `FlickingOptions` and you can define the movement of Flicking inside Flicking.

If `nested` is set to true, the parent Flicking will start moving if the child Flicking reaches the start/end of the panels.
If `nested` is set to false, parent Flicking does not move when user swipes child Flicking.

nesteddemov2
nested2

Also added examples of nested option applied to Options page and Demos page of documentation.

### DIFF
--- a/docs/src/css/custom.css
+++ b/docs/src/css/custom.css
@@ -168,3 +168,21 @@ article a {
   width: auto;
   height: 100px;
 }
+
+.nested-wide .flicking-panel {
+  width: 100%;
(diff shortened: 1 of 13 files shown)


qpitlove (approved):
LGTM

daybrush (approved):
LGTM

WoodNeck (approved):
LGTM

Issue #673, opened 2022-05-16
Flicking components with horizontal drag block vertical drag


gbyesiltas:
## Description
Flicking components with horizontal drag block vertical drag on mobile. I believe this happened after 4.6.3 or 4.7.0

## Steps to check or reproduce
Go to the [demo on the flicking website](https://naver.github.io/egjs-flicking/Demos) on mobile (or on a mobile viewport on the devtools) and try to drag vertically by starting your drag on the flicking component. It won't cause a vertical scroll.

malangfox:
Hello @gbyesiltas.
We have checked the inappropriate behavior you left behind.
This issue will be fixed in the next patch. We will notify you after the problem you reported is corrected.
Thank you for reporting!

gbyesiltas:
> Hello @gbyesiltas.
> We have checked the inappropriate behavior you left behind.
> This issue will be fixed in the next patch. We will notify you after the problem you reported is corrected.
> Thank you for reporting!

Thank you! :)

WoodNeck:
@hyeon-9yu 
I believe this happened after [4.7.0](https://github.com/naver/egjs-flicking/releases/tag/4.7.0), so you can either:
- use 4.6.x instead
```sh
npm i @egjs/flicking@~4.6.0
```
- Add the following CSS to the `.flicking-viewport` element
```css
.flicking-viewport {
  touch-action: pan-y !important;
}
```

Fix will be made after https://github.com/naver/egjs-axes/pull/185

hyeon-9yu:
> @hyeon-9yu I believe this happened after [4.7.0](https://github.com/naver/egjs-flicking/releases/tag/4.7.0), so you can either:
> 
> * use 4.6.x instead
> 
> ```shell
> npm i @egjs/flicking@~4.6.0
> ```
> 
> * Add the following CSS to the `.flicking-viewport` element
> 
> ```css
> .flicking-viewport {
>   touch-action: pan-y !important;
> }
> ```
> 
> Fix will be made after [naver/egjs-axes#185](https://github.com/naver/egjs-axes/pull/185)

Thank you for help!
I've been using the 4.4.3 version of react-flicking since before, so I don't know why it's happening.
And I tried the method you provided, but it seems that the same issue still occurs in IOS.

WoodNeck:
@hyeon-9yu
Did you try downgrading the version to 4.6.x?
If not, please try that way instead. We'll fix this asap in the next version :)

malangfox:
[...]

PR #675, opened 2022-05-19
fix: use appropriate touch-action for element

malangfox:
## Issue
#673 

## Details
It uses the 3.2.0 version of the Axes and sets appropriate touch-action according to the drag direction of Flicking.
For horizontal flicking, touch-action is set to pan-y. And for vertical flicking, touch-action is set to pan-x.


daybrush (approved):
LGTM

WoodNeck (approved):
LGTM

PR #682, opened 2022-05-27
fix: wrong camera offset on adaptive resize

WoodNeck:
## Details
- This fixes Flicking having the wrong camera offset after calling `resize()` in some cases.
- This also fixes Flicking doesn't update height after `resize` when `adaptive: true`


daybrush (approved):
LGTM

malangfox (approved):
LGTM

PR #684, opened 2022-05-30
feat: add changeOnHold

WoodNeck:
## Details
This adds a new option "changeOnHold", which triggers `changed` event on mouse/touch hold while animating.

### DIFF
--- a/docs/src/pages/Options.mdx
+++ b/docs/src/pages/Options.mdx
@@ -1050,6 +1050,45 @@ Automatically call **[disableInput()](docs/api/Flicking#disableInput)** on initi
   </ColumnItem>
 </Columns>
 
+### changeOnHold
+Change active panel index on mouse/touch hold while animating.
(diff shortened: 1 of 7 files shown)


docs/src/pages/Options.mdx
```diff
@@ -1050,6 +1050,45 @@ Automatically call **[disableInput()](docs/api/Flicking#disableInput)** on initi
   </ColumnItem>
 </Columns>
 
+### changeOnHold
+Change active panel index on mouse/touch hold while animating.
```
daybrush, line 1054: Is it only necessary for hold? It seems like a good feature to change the activePanel even when not on hold.

daybrush (approved):
LGTM

malangfox (approved):
LGTM

docs/src/pages/Options.mdx
WoodNeck: Umm... like when? Can you show me some example cases?

PR #686, opened 2022-06-03
refactor: add union type to ViewportSlot

RudyBlack:
## Details
Sometimes ViewportSlot has only one children such as Plugin Pagination. 
So I think ViewportSlot children type must be React.ReactElement[] or React.ReactElement.

image

### DIFF
--- a/packages/react-flicking/src/react-flicking/ViewportSlot.tsx
+++ b/packages/react-flicking/src/react-flicking/ViewportSlot.tsx
@@ -4,6 +4,6 @@
  */
 import React from "react";
 
-const ViewportSlot = React.memo((props: { children?: React.ReactElement[] }) => <>{props.children}</>);
+const ViewportSlot = React.memo((props: { children?: React.ReactElement[] | React.ReactElement }) => <>{props.children}</>);
 
 export default ViewportSlot;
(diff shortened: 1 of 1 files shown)


WoodNeck:
Hello @RudyBlack!
I have a few questions.
1. Did it trigger errors before?
2. Original definition of the `children` prop in React is like this:
```js
readonly props: Readonly<P> & Readonly<{ children?: ReactNode | undefined }>;
``` 
Shouldn't the types be like that to make it strict?

daybrush (approved):
LGTM

RudyBlack:
> Hello @RudyBlack! I have a few questions.
> 
> 1. Did it trigger errors before?
> 2. Original definition of the `children` prop in React is like this:
> 
> ```js
> readonly props: Readonly<P> & Readonly<{ children?: ReactNode | undefined }>;
> ```
> 
> Shouldn't the types be like that to make it strict?


1. Trigger type error this

Screen Shot 2022-06-03 at 5 37 07 PM
Screen Shot 2022-06-03 at 5 36 39 PM
.

2. Oh, That is more better!

WoodNeck:
@RudyBlack 
Thanks for the answer, could you change the type to what I've suggested in that case?

RudyBlack:
ReactNode type is include `undefined`. so I just change ReactNode type.

WoodNeck (approved):
LGTM

malangfox (approved):

PR #687, opened 2022-06-03
fix: prevent click works even when input is disabled

malangfox:
## Details
Fixed the issue that next click did not work when moving the Flicking using `moveTo` after `disableInput`.

`disableInput` disables PanInput, so setting `dragged` to `false` when `hold` is made in `PanInput` no longer occurs.
If coordinate change is made in Axes using `moveTo` while `PanInput` is disabled, `dragged` becomes `true`, but `hold` does not occur in PanInput at the next click.

[Example of onClick not working after using moveTo by button while input is disabled.](https://codepen.io/malangfox/pen/zYRawpb) To check this error, click a Panel once, and click Next button, and click a Panel.

Since `PanInput` is used in Flicking for dragging gesture, we should set `dragged` to `true` only when `PanInput` is enabled.

### DIFF
--- a/src/control/AxesController.ts
+++ b/src/control/AxesController.ts
@@ -368,7 +368,7 @@ class AxesController {
   };
 
   private _onAxesChange = () => {
-    this._dragged = true;
+    this._dragged = !!this._panInput?.isEnabled() && true;
   };
 
   private _preventClickWhenDragged = (e: MouseEvent) => {
--- a/src/control/AxesController.ts
+++ b/src/control/AxesController.ts
@@ -368,7 +368,7 @@ class AxesController {
   };
 
   private _onAxesChange = () => {
-    this._dragged = !!this._panInput?.isEnabled() && true;
+    this._dragged = !!this._panInput?.isEnabled();
   };
 
   private _preventClickWhenDragged = (e: MouseEvent) => {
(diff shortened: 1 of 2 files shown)


src/control/AxesController.ts
```diff
@@ -368,7 +368,7 @@ class AxesController {
   };
 
   private _onAxesChange = () => {
-    this._dragged = true;
+    this._dragged = !!this._panInput?.isEnabled() && true;
```
WoodNeck, line 371: How about checking it with properties like `this._state.holding` or `isTrusted` property of the Axes event?
Adding more inputs like WheelInput can break this.

malangfox: Since we are blocking the click event through the `preventClickOnDrag` option,
I don't think there's any need to prevent click event when we move Panel with Wheel or MoveKey etc.

WoodNeck: That makes sense 👍

WoodNeck (approved):
LGTM

src/control/AxesController.ts
malangfox: Aside from that, `&& true` was removed because it looked unnecessary.

daybrush (approved):
LGTM

PR #701, opened 2022-07-07
fix: panelsPerView not working with adaptive

malangfox:
## Details
If you use the `panelsPerView` option in horizontal flicking, the height of all panels is also internally made to be the same as the first panel.

This causes invalid behavior at the `adaptive` option that resizes the viewport via the Panel’s height.

[Demo - adaptive option not working](https://codesandbox.io/s/adaptive-flicking-ilvxyr?file=/src/App.vue)

Also, when using the `bound` option with `panelsPerView`, the incorrectly calculated size will cause an error when moving the Panel.

[Demo - panelsPerView option causing error](https://codesandbox.io/s/panelsperview-flicking-error-k9ef7h?file=/src/App.vue)

This fixes both by modifying the behavior to resize all Panels instead of just the first one and using it for resizing all.

### DIFF
--- a/src/renderer/Renderer.ts
+++ b/src/renderer/Renderer.ts
@@ -494,17 +494,19 @@ abstract class Renderer {
     const panelSizeObj = flicking.horizontal
       ? { width: panelSize }
       : { height: panelSize };
-    const firstPanelSizeObj = {
-      size: panelSize,
-      height: referencePanel.height,
-      margin: referencePanel.margin
-    };
 
-    if (!flicking.noPanelStyleOverride) {
+    if (flicking.noPanelStyleOverride) {
+      const firstPanelSizeObj = {
+        size: panelSize,
+        height: referencePanel.height,
+        margin: referencePanel.margin
+      };
+      flicking.panels.forEach(panel => panel.resize(firstPanelSizeObj));
+    } else {
       this._strategy.updatePanelSizes(flicking, panelSizeObj);
+      flicking.panels.forEach(panel => panel.resize());
     }
 
-    flicking.panels.forEach(panel => panel.resize(firstPanelSizeObj));
   }
 
   protected _removeAllChildsFromCamera() {
--- a/test/manual/adaptive.html
+++ b/test/manual/adaptive.html
@@ -1,0 +1,80 @@
+<!DOCTYPE html>
+<html lang="en">
+<style>
+  * {
+    box-sizing: border-box;
+  }
+
+  #app {
+    font-family: "Avenir", Helvetica, Arial, sans-serif;
+    -webkit-font-smoothing: antialiased;
+    -moz-osx-font-smoothing: grayscale;
+    text-align: center;
+    color: #2c3e50;
+    margin-top: 60px;
+    background: #F2F4F7;
+    width: 600px;
+    margin-left: auto;
+    margin-right: auto;
+  }
+
+  .flicking-viewport {
+  transition: height 500ms;
+}
+  .panel {
+    background: #D7E6FF;
+    padding: 40px;
+    margin-right: 20px;
+  }
+
+.long {
+  height: 300px;
+}
+
+.short {
+  height: 150px;
+}
+
+.flicking-container {
+  padding: 10px 0;
+}
+
+</style>
+
+<head>
+  <meta name="viewport"
+    content="width=device-width,initial-scale=1.0,maximum-scale=1.0,minimum-scale=1.0,user-scalable=no">
+  <meta charset="utf-8">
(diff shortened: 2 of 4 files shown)


src/renderer/Renderer.ts
```diff
@@ -494,17 +494,19 @@ abstract class Renderer {
...
+    if (flicking.noPanelStyleOverride) {
+      const firstPanelSizeObj = {
+        size: panelSize,
+        height: referencePanel.height,
+        margin: referencePanel.margin
+      };
+      flicking.panels.forEach(panel => panel.resize(firstPanelSizeObj));
+    } else {
       this._strategy.updatePanelSizes(flicking, panelSizeObj);
+      flicking.panels.forEach(panel => panel.resize());
```
WoodNeck, line 507: As `panelsPerView` restricts all panels to have the same size, we can improve the initialization performance by calculating the panel size of only one element.

If you resize all the panels like this, panels will calculate the sizes of all elements.
In that case, `panelsPerView` has no meaning as the user can restrict the panel size not by option but by CSS.

Please change this to calculate the `height` of the element only, also consider the option `horizontal` while doing that.

WoodNeck (changes requested):

src/core/panel/Panel.ts
```diff
@@ -321,7 +321,13 @@ class Panel {
...
       this._size = cached.size;
       this._margin = { ...cached.margin };
-      this._height = cached.height;
+      this._height = horizontal ? getElementSize({
+        el,
+        horizontal: false,
+        useFractionalSize,
+        useOffset: true,
+        style: getStyle(el)
+      }) : cached.height;
```
WoodNeck, line 330: This affects all other cases not using `panelsPerView`, this should be reverted.
Or, you should check whether to calculate the height by `cached.height` not `horizontal`

src/renderer/Renderer.ts
malangfox: Considering that calculating the height through every element style has a direct impact on performance,
it seems appropriate to use a cached panel as much as possible, but change it to calculate the height only when it is horizontal.

src/core/panel/Panel.ts
```diff
@@ -321,7 +321,13 @@ class Panel {
     if (cached) {
       this._size = cached.size;
       this._margin = { ...cached.margin };
-      this._height = cached.height;
+      this._height = horizontal ? getElementSize({
```
daybrush, line 324: Is it okay to check only when it is horizontal?

daybrush (approved):
LGTM

src/core/panel/Panel.ts
WoodNeck: `horiztontal` I meant on the previous comment was to update height only when using `panelsPerView` and `horizontal: true` as when `horitzontal: false` all panels will use all same heights.

WoodNeck (changes requested):

src/core/panel/Panel.ts
malangfox: Currently it seems `cached` is passed to `resize` only via `updatePanelSizeByGrid` when `panelsPerView` is greater than 0.
However considering the purpose of `resize`, it seems better to determine whether to include `height` in `firstPanelSizeObj` in `updatePanelSizeByGrid`.

src/renderer/Renderer.ts
```diff
@@ -496,8 +496,8 @@ abstract class Renderer {
       : { height: panelSize };
     const firstPanelSizeObj = {
       size: panelSize,
-      height: referencePanel.height,
-      margin: referencePanel.margin
+      margin: referencePanel.margin,
+      ...(!flicking.horizontal && { height: referencePanel.height} )
```
WoodNeck, line 500: Wrong spaces here
```js
 ...(!flicking.horizontal && { height: referencePanel.height })
```

src/core/panel/Panel.ts
```diff
@@ -321,13 +321,13 @@ class Panel {
     if (cached) {
       this._size = cached.size;
       this._margin = { ...cached.margin };
-      this._height = horizontal ? getElementSize({
+      this._height = cached.height ? cached.height : getElementSize({
```
WoodNeck, line 324: This can result in a recalculation of height when `cached.height` equals 0.
Please use `??` instead

malangfox: I think `??` may have issues with Flicking's browser compatibility.
There is "IE 9+(With polyfill), 11+ for Angular & Svelte" in supported browsers, [but it looks like we won't be able to use ?? in these versions.](https://caniuse.com/?search=%3F%3F)

Issue #702, opened 2022-07-07
SnapControl throws an error when bound is true under certain circumstances.


malangfox:
## Description
![image](https://user-images.githubusercontent.com/13797320/177722552-cdb6f24d-45ad-4218-af89-b3d8207c4518.png)

In some circumstances when `bound` is set to true, Flicking throws an error and doesn't seem to snap properly.

## Steps to check or reproduce
https://codepen.io/malangfox/pen/bGvEarp

PR #703, opened 2022-07-08
fix: bound option causing error when using with specific panel width

malangfox:
## Issue
#702 

## Details
When the width of each panel is less than 33% of the viewport, using the SnapControl and `bound` options together will cause an error.

We use [`findActiveAnchor`](https://github.com/naver/egjs-flicking/blob/master/src/camera/Camera.ts#L394-L399) in `moveToPosition` of SnapControl.
However ` find(this._anchors, anchor => anchor.panel.index === activeIndex)` can return `null` depending on the panel index of the anchors. This throws `POSITION_NOT_REACHABLE` error that makes SnapControl not working.

By default, [Flicking's `defaultIndex` is 0.](https://github.com/naver/egjs-flicking/blob/master/src/Flicking.ts#L799)  Also Flicking [sets the initial panel](https://github.com/naver/egjs-flicking/blob/master/src/Flicking.ts#L1505) to this 0th panel.
When the width of each panel is less than 33% of the viewport, the anchor with the 0th panel may not exist.

image

This makes the center of the viewport to the anchor of the 0th Panel. This situation does not reproduce an error.

image

This makes the center of the viewport to the anchor of the first Panel. This situation causes error because the panel with the first anchor point has an index of 1, so there is no anchor with the 0th panel.

The 0th panel may never be selected depending on the width of the panels. We should apply this situation to Flicking's initialization process as well.

This changes `initialPanel` if the Panel index of the `nearestAnchor` is different from the index of the `initialPanel` when Flicking is initialized.

### DIFF
--- a/src/Flicking.ts
+++ b/src/Flicking.ts
@@ -1502,11 +1502,12 @@ class Flicking extends Component<FlickingEvents> {
     const renderer = this._renderer;
     const control = this._control;
     const camera = this._camera;
-    const initialPanel = renderer.getPanel(this._defaultIndex) || renderer.getPanel(0);
+    const defaultPanel = renderer.getPanel(this._defaultIndex) || renderer.getPanel(0);
 
-    if (!initialPanel) return;
+    if (!defaultPanel) return;
 
-    const nearestAnchor = camera.findNearestAnchor(initialPanel.position);
+    const nearestAnchor = camera.findNearestAnchor(defaultPanel.position);
+    const initialPanel = (nearestAnchor && defaultPanel.index !== nearestAnchor?.panel.index) ? nearestAnchor.panel : defaultPanel;
     control.setActive(initialPanel, null, false);
 
     if (!nearestAnchor) {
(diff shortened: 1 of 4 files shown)


src/Flicking.ts
```diff
@@ -1502,11 +1502,12 @@ class Flicking extends Component<FlickingEvents> {
...
     const camera = this._camera;
-    const initialPanel = renderer.getPanel(this._defaultIndex) || renderer.getPanel(0);
+    const defaultPanel = renderer.getPanel(this._defaultIndex) || renderer.getPanel(0);
 
-    if (!initialPanel) return;
+    if (!defaultPanel) return;
 
-    const nearestAnchor = camera.findNearestAnchor(initialPanel.position);
+    const nearestAnchor = camera.findNearestAnchor(defaultPanel.position);
+    const initialPanel = (nearestAnchor && defaultPanel.index !== nearestAnchor?.panel.index) ? nearestAnchor.panel : defaultPanel;
```
WoodNeck, line 1510: Is `?` needed after `nearestAnchor`?
I mean, shouldn't the condition be
```js
(nearestAnchor && defaultPanel.index !== nearestAnchor.panel.index)
```

WoodNeck (approved):
LGTM

src/Flicking.ts
malangfox: Your comment is right. The unnecessary ? has been removed.

daybrush (approved):
LGTM

PR #711, opened 2022-07-26
fix: fix error when onChange occur during holding state

malangfox:
## Issue
#709 

## Details
If `onChange` occurs while Flicking is in the holding state, we check the `offsetX` or `offsetY` of the native event that caused the change.
```ts
    const inputEvent = axesEvent.inputEvent as { offsetX: number; offsetY: number };
    const offset = flicking.horizontal
      ? inputEvent.offsetX
      : inputEvent.offsetY;
```

However, there is no native event when we try to move the holding state Flicking using the method like `moveTo`. This will cause an error `Cannot read properties of null (reading 'offsetX')`.
[You can check the error in this demo.](https://codepen.io/malangfox/pen/NWYvQKW)

This changes `onChange` to check the direction using `animatingContext` instead of the native event when Flicking moves in a way other than a native event.

### DIFF
--- a/src/Flicking.ts
+++ b/src/Flicking.ts
@@ -1084,6 +1084,10 @@ class Flicking extends Component<FlickingEvents> {
       return Promise.reject(new FlickingError(ERROR.MESSAGE.ANIMATION_ALREADY_PLAYING, ERROR.CODE.ANIMATION_ALREADY_PLAYING));
     }
 
+    if (this._control.holding) {
+      this._control.controller.release();
+    }
+
     return this._control.moveToPanel(panel, {
       duration,
       direction
--- a/src/control/states/HoldingState.ts
+++ b/src/control/states/HoldingState.ts
@@ -36,14 +36,17 @@ class HoldingState extends State {
 
   public onChange(ctx: Parameters<State["onChange"]>[0]): void {
     const { flicking, axesEvent, transitTo } = ctx;
-    const animatingContext = flicking.control.controller.animatingContext;
+
     const inputEvent = axesEvent.inputEvent as { offsetX: number; offsetY: number };
-    const offset = flicking.horizontal ? inputEvent?.offsetX : inputEvent?.offsetY;
+
+    const offset = flicking.horizontal
+      ? inputEvent.offsetX
+      : inputEvent.offsetY;
 
     const moveStartEvent = new ComponentEvent(EVENTS.MOVE_START, {
       isTrusted: axesEvent.isTrusted,
       holding: this.holding,
-      direction: inputEvent ? getDirection(0, -offset) : getDirection(animatingContext.start, animatingContext.end),
+      direction: getDirection(0, -offset),
       axesEvent
     });
     flicking.trigger(moveStartEvent);
--- a/src/control/states/HoldingState.ts
+++ b/src/control/states/HoldingState.ts
@@ -39,9 +39,11 @@ class HoldingState extends State {
 
     const inputEvent = axesEvent.inputEvent as { offsetX: number; offsetY: number };
 
-    const offset = flicking.horizontal
-      ? inputEvent.offsetX
-      : inputEvent.offsetY;
+    const offset = inputEvent
+      ? flicking.horizontal
+        ? inputEvent.offsetX
+        : inputEvent.offsetY
+      : 0;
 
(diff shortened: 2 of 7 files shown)


src/control/states/HoldingState.ts
```diff
@@ -36,17 +36,14 @@ class HoldingState extends State {
...
-      ? inputEvent.offsetX
-      : inputEvent.offsetY;
+    const offset = flicking.horizontal ? inputEvent?.offsetX : inputEvent?.offsetY;
 
     const moveStartEvent = new ComponentEvent(EVENTS.MOVE_START, {
       isTrusted: axesEvent.isTrusted,
       holding: this.holding,
-      direction: getDirection(0, -offset),
+      direction: inputEvent ? getDirection(0, -offset) : getDirection(animatingContext.start, animatingContext.end),
       axesEvent
```
WoodNeck, line 47: This will end up Flicking being in `DraggingState`, but it should actually be in `AnimatingState`.
You should rather change Flicking's state to `IdleState` before calling moveTo / moveToPanel.

WoodNeck (changes requested):

src/control/states/HoldingState.ts
malangfox: In the idle state, we change state to `AnimatingState` when onChange occurs.
Does this assume that there is no other way to fire change event in idle other than starting animation through the methods?

If so, it seems to be possible to set the `DraggingState` only when there is a native event as user input in the holding state, and if there is no native event, we can set it to `AnimatingState` like in the idle.
Can it be appropriate to define the next state like this?
`const nextState = inputEvent ? STATE_TYPE.DRAGGING : STATE_TYPE.ANIMATING;`

WoodNeck: The current state flow is:
- By user input: IDLE -> HOLDING -> DRAGGING -> ANIMATING -> IDLE (or IDLE -> HOLDING -> IDLE)
- By method call: IDLE -> ANIMATING -> IDLE

I think it's better to cancel the current input sequence on the method call than adding the new state flow like that.
Like, in the demo you provided(https://codepen.io/malangfox/pen/NWYvQKW) it triggers `Animation is already playing.` error if `moveTo` is called when Flicking's in `DraggingState`. fixing only the `HoldingState` doesn't fix that.

Also, as this fix doesn't care about previous inputs after changing it to `AnimatingState`, that will end up like this:


You should cancel the previous inputs with Axes if the Flicking's `control.holding` is true, change the current state back to idle, and handle the moveTo animation like normal.

malangfox:
@WoodNeck 

When the method `moveTo` is called with user input is in the holding state, Now we can stop the input by calling the `release` of `PanInput`.
Since this newly applied in `moveTo`, `holdEnd` event fires in above case and it changes the status of Flicking to `IDLE`.

As a result, `moveTo` changes the status of Flicking in the following order when user input is holding:
IDLE -> HOLDING -> IDLE -> ANIMATING -> IDLE

Not only that, we can now improve to use the `moveTo` method even when in `DRAGGING` state via user input.
However, in this case in current implementation, It throws an error `Animation is already playing.` and is rejected, so this will be further discussion.

WoodNeck (approved):
LGTM
I guess you can improve Flicking by making it can change the animation state whenever the user wants, like you've said.

daybrush (approved):
LGTM

src/Flicking.ts
```diff
@@ -1084,6 +1084,10 @@ class Flicking extends Component<FlickingEvents> {
       return Promise.reject(new FlickingError(ERROR.MESSAGE.ANIMATION_ALREADY_PLAYING, ERROR.CODE.ANIMATION_ALREADY_PLAYING));
     }
 
+    if (this._control.holding) {
```
daybrush, line 1087: Is it correct to do moveTo while dragging?

malangfox: This fixes an issue that causes `moveTo` to throw an error when the input is holding on Flicking.

However, `control.animating` is false at `HOLDING` state, but becomes true in `DRAGGING` state.
When in DRAGGING state, `moveTo` throws an `ANIMATION_ALREADY_PLAYING` error above. 

It has been changed to prevent unhandled error by releasing user input using `release`, Should we block `moveTo` while in the holding state?

PR #717, opened 2022-08-10
fix: fix updated state not applied in event handler functions in React

malangfox:
## Issue
#716 

## Details
This fixes the error that the event handler registered in react-flicking does not applies the updated state.
If we use the state declared be `useState` inside the event handler, the most recently updated state is not reflected correctly.

[In this demo, you can compare the results when adding a log by pressing a button and executing the same function as an event handler by moving Flicking.](https://codesandbox.io/s/egjs-flicking-v4-on-move-end-891kwx)
If this same function `holdStart` is executed as an event handler for Flicking, logs are recognized as an empty array in its initial state and the array will become empty.

This has been fixed by using event handlers with the correct scope.

### DIFF
--- a/packages/react-flicking/src/react-flicking/Flicking.tsx
+++ b/packages/react-flicking/src/react-flicking/Flicking.tsx
@@ -198,7 +198,6 @@ class Flicking extends React.Component<Partial<FlickingProps & FlickingOptions>>
 
   private _bindEvents() {
     const flicking = this._vanillaFlicking!;
-    const props = this.props as Required<FlickingProps>;
 
     Object.keys(EVENTS).forEach((eventKey: keyof typeof EVENTS) => {
       const eventName = EVENTS[eventKey];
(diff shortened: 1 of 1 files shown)


packages/react-flicking/src/react-flicking/Flicking.tsx
```diff
@@ -207,7 +206,10 @@ class Flicking extends React.Component<Partial<FlickingProps & FlickingOptions>>
       flicking.on(eventName, e => {
         e.currentTarget = this;
 
-        props[propName](e);
+        const evtHandler = this.props[propName];
+        if (evtHandler) {
+          evtHandler(e);
+        }
```
WoodNeck, line 212: As all Flicking event handlers have a default value, checking existence (`if (eventHandler)`) can be omitted.
- https://github.com/naver/egjs-flicking/blob/e6a393d3f061141091bac9f2b156970b43e6701e/packages/react-flicking/src/react-flicking/consts.ts#L33-L49

WoodNeck (approved):
LGTM

PR #718, opened 2022-08-11
fix: fix error when in exception when change occur without native event

malangfox:
## Issue
#711 , https://github.com/naver/egjs-axes/issues/197

## Details
We previously checked an error that occurred because the native event did not exist when useing `moveTo` method fires the change event in the `HOLDING` state.
However, in addition to this case, it seems that the change event without native event while in the `HOLDING` state may occur.

This case is uncommon, but using some ways such as calling `setTo` on the Axes instance other than using Flicking's `moveTo` can be this case. This also add this case in our unit tests.

In other unexpected cases may exist while in `HOLDING` state. For this reason, when there is no native event, the direction is determined from the `animatingContext`.

### DIFF
--- a/src/control/states/HoldingState.ts
+++ b/src/control/states/HoldingState.ts
@@ -36,17 +36,15 @@ class HoldingState extends State {
 
   public onChange(ctx: Parameters<State["onChange"]>[0]): void {
     const { flicking, axesEvent, transitTo } = ctx;
-
+    const animatingContext = flicking.control.controller.animatingContext;
     const inputEvent = axesEvent.inputEvent as { offsetX: number; offsetY: number };
+    const offset = flicking.horizontal ? inputEvent?.offsetX : inputEvent?.offsetY;
 
-    const offset = flicking.horizontal
-      ? inputEvent.offsetX
-      : inputEvent.offsetY;
 
     const moveStartEvent = new ComponentEvent(EVENTS.MOVE_START, {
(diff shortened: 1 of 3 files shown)


src/control/states/HoldingState.ts
```diff
@@ -36,17 +36,15 @@ class HoldingState extends State {
...
 
-    const offset = flicking.horizontal
-      ? inputEvent.offsetX
-      : inputEvent.offsetY;
 
     const moveStartEvent = new ComponentEvent(EVENTS.MOVE_START, {
       isTrusted: axesEvent.isTrusted,
       holding: this.holding,
-      direction: getDirection(0, -offset),
+      direction: inputEvent ? getDirection(0, -offset) : getDirection(animatingContext.start, animatingContext.end),
```
WoodNeck, line 47: The fix may be okay, but I'm a bit concerned about triggered events.
In the past PR, you've fixed the way that triggers the `holdEnd` event, while this doesn't.
So, for the two same use cases, events are triggered in a different way.
1. `holdStart` -> `holdEnd` -> `moveStart` (previous PR)
2. `holdStart` -> `moveStart` (current PR)

Please check the actual cause why the previous PR doesn't resolve the issue, and fix this PR in more proper way.

WoodNeck (changes requested):

daybrush (approved):
LGTM

PR #728, opened 2022-09-07
feat: add updateAnimation and stopAnimation methods

malangfox:
## Issue
#725 

## Details
This provides methods to change the state of the animation being played.
`updateAnimation` can change the destination, time and direction of the animation being played.
`stopAnimation` triggers the `moveEnd` event and stops the animation in progress.
`change` event will not trigger for animations stopped with `stopAnimation`.


daybrush (approved):
LGTM

src/control/AxesController.ts
```diff
@@ -214,6 +214,39 @@ class AxesController {
...
+   * @return {this}
+   */
+  public updateAnimation(position: number, duration?: number): this {
+    this._animatingContext = {
+      ...this._animatingContext,
+      end: position
+    };
+    this._axes?.updateAnimation({
+      destPos: { [AXES.POSITION_KEY]: position },
+      duration
```
WoodNeck, line 232: As changing `duration` in Flicking does not directly change Axes's option `duration`, you should use Flicking's default duration as a fallback.

```js
duration: duration ?? this._flicking.duration
```

WoodNeck (approved):
LGTM

src/control/AxesController.ts
malangfox: `updateAnimation` of the Axes keeps the duration of the current animation when the [duration does not exist.](https://github.com/naver/egjs-axes/blob/master/packages/axes/src/animation/EasingManager.ts#L29)
However, I found that replacing the || with ?? in the code linked above works correctly when the new duration is 0.
This will be fixed in Axes' recent PR.

PR #734, opened 2022-09-28
fix: fix panels could not be updated inside hooks

malangfox:
## Issue
#719 

## Details
This fixes the bug in react-flicking that the updated panels are not applied when the list of panels are updated in hooks.

### DIFF
--- a/packages/react-flicking/src/react-flicking/Flicking.tsx
+++ b/packages/react-flicking/src/react-flicking/Flicking.tsx
@@ -26,12 +26,6 @@ import NonStrictPanel from "./NonStrictPanel";
 import ViewportSlot from "./ViewportSlot";
 import ReactElementProvider from "./ReactElementProvider";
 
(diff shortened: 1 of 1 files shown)


WoodNeck (approved):
LGTM

packages/react-flicking/src/react-flicking/Flicking.tsx
```diff
@@ -92,31 +87,40 @@ class Flicking extends React.Component<Partial<FlickingProps & FlickingOptions>>
...
       }
     }
 
     return false;
   }
 
+  public beforeRender() {
+    const vanillaFlicking = this._vanillaFlicking;
+    const nextProps = this.props;
+    const prevProps = this._prevProps;
```
daybrush, line 111: You can get prevProps from componentDIdupdate.

daybrush (approved):
LGTM

PR #735, opened 2022-09-30
feat: add renderOnSamekey for react-flicking

WoodNeck:
## Details
This adds a new prop `renderOnSameKey` for react-flicking, which can disable Flicking's behavior which it doesn't render when the prop `children` has the same length & keys.

### DIFF
--- a/docs/docs/tutorials/quick-start.mdx
+++ b/docs/docs/tutorials/quick-start.mdx
@@ -91,6 +91,15 @@ export default () => (
 )
 ```
 
+<h3>React exclusive options</h3>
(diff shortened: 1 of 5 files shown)


malangfox (approved):
LGTM

daybrush (approved):
LGTM

malangfox:
#736 
It has been reported that when `_panel` of  react-flicking is updated, it is not applied correctly in the original flicking's `panel` array.
This part may also be fixed with this PR, updating original `panel` array when `_panel` is updated.

Issue #740, opened 2022-10-27
phantom dependencies issue with react-flicking


daybrush:
## Description

* package.json
```js
"dependencies": {
    "@egjs/flicking": "~4.10.1",
    "@egjs/list-differ": "^1.0.0"
  },
```
https://github.com/naver/egjs-flicking/blob/master/packages/react-flicking/src/react-flicking/Flicking.tsx
```js
import Component from "@egjs/component";
```

PR #741, opened 2022-10-27
fix: fix component dependency for frameworks

daybrush:
## Issue

#740

### DIFF
--- a/packages/ngx-flicking/projects/ngx-flicking/ng-package.json
+++ b/packages/ngx-flicking/projects/ngx-flicking/ng-package.json
@@ -11,6 +11,7 @@
   },
   "whitelistedNonPeerDependencies": [
     "@egjs/flicking",
-    "@egjs/list-differ"
+    "@egjs/list-differ",
+    "@egjs/component"
   ]
 }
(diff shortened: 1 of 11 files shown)


WoodNeck (changes requested):
I think `ngx-flicking` and `vue-flicking` should be excluded from this PR, as they don't directly import `@egjs/component` in their code.
`ngx-flicking` actually imports `ComponentEvent`, but as it's just importing type from the component, I hope that's okay.
Or, we can  try changing it explicitly like this:
```ts
import type { ComponentEvent } from "@egjs/component";
```

Anyway, I wonder in which env the issue occurs.
As `@egjs/component` is resolved correctly in the same `node_moudles` directory in most cases.

daybrush:
In most cases there is no problem.

There was a report from yarn berry that it was installed as follows.
@egjs/react-flicking
@egjs/flicking
@egjs/flicking/node_modules/@egjs/component


There may be no problems here, but phantom dependencies may have version issues.

WoodNeck (approved):
LGTM

I think we may have to update the minor version of the framework packages next time, when we upgrade the `@egjs/component` to v4.

malangfox (approved):
LGTM

PR #743, opened 2022-11-01
fix: some setters doesn't update UI

### LINKED ISSUE #742: [react-flicking] PanelsPerView does not update
## Description

- On react, changing the `panelsPerView` property during runtime does not affect the corresponding `panelsPerView` property on the object that backs the `<Flicker />` component.
- On the other hand, preparing a `Ref` and setting `flicker.current.panelsPerView` resizes the component automatically.
- The property itself works if you keep it hardcoded, it just happens that the underlying state of the component doesn't follow the property.

- 1) Set up a `create-react-app` project 
- 2) Replace the `<App/>` tag with this `<FT />` tag (rename the import on `index.js` as well):
```
import * as React from 'react'
import { useRef } from 'react'

import Flicking from '@egjs/react-flicking'
import "@egjs/react-flicking/dist/flicking.css"

function FT(){
	const sample_panel = [1,2,3];
	const [nPans, setNPans] = React.useState(2)

	const theFlicking = useRef(null)

	return (
		<div style={{width:"100%",height:"100%"}}>
			<button style={{
				width:"270px", 
				backgroundColor:"#335FE1", 
				textAlign:"center",
				color:"#E9DAE2",
				fontWeight:"bold",
				fontSize:"60px" }}
 
				//change panels-per-view on clicking the button.
				onClick={() => {
					console.log("toggle clicked.")

					//using temporary variable for new panel count
					let marau=1
					if(theFlicking.current.panelsPerView==1){
						marau = 2
					}
					else{
						marau = 1
					}
					//none of these change the panels.
					theFlicking.current.forceUpdate()
					theFlicking.current.resize()

					//setting the variable directly on the class works.
					theFlicking.current.panelsPerView=marau

					//reset index directly.
					theFlicking.current.moveTo(0)

					console.log("toggle button clicked, Flicking.panelsPerView is now " + theFlicking.current.panelsPerView)
				}}>
		TOGGLE</button>
				<h5>panelsPerView react property is now {nPans}</h5>
				<Flicking
					//React ref to call methods and set properties.
					ref={theFlicking}

					//Changing this property on the component does not change the corresponding property on the javascript object.
					panelsPerView={nPans}

					align="prev"
					onChanged={() => {
						if (nPans==2){
							setNPans(1)
						}
						else{
							setNPans(2)
						}
					}}
				>
					{sample_panel.map((e, i) => 
						<div className="flicking-panel" key={i} ><div width="100%"><h2>{e}</h2></div></div>)
					}
				</Flicking>
		</div>
	)
}

export default FT;
```
If you set things up correctly, hitting up the development server (usually `localhost:3000`) shows a test page.
- 3) Click the "TOGGLE" button at the top of the page. It sets the property directly on the Flicker object, which then resizes the panels to match.
- 4) Flick on the Flicker. Everytime the index changes, the `nPans` state variable toggles between 1 and 2, thus changing the `panelsPerView` property on the JSX element, but the Flicker object does not resize the panels to match.

## some additional details
- calling `resize()`, `forceUpdate()`, one or both, they don't trigger panel resize.

WoodNeck:
## Issue
#742 

## Details
This fixes #742


packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
```diff
@@ -200,9 +201,20 @@ export class NgxFlickingComponent extends FlickingInterface
...
-    );
+    this._ngZone.runOutsideAngular(() => {
+      void flicking.renderer.forceRenderAllPanels();
+    });
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const { virtual, ...newOptions } = this.options;
+
+    // Omit 'virtual', as it can't have any setter
+    for (const key in newOptions) {
```
daybrush, line 212: The prop diff result is in the first parameter, try this one.

https://angular.io/api/core/OnChanges#ngOnChanges

daybrush (approved):
LGTM

packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
WoodNeck: Well, there's no meaning on using that as Flicking bounds all options to prop `options`
image

Have to check every props inside `options` in either way.

malangfox (approved):
LGTM

As the gap option removed in v4, I checked that this PR removes the gap demo of react-flicking.
Can we remove the associated css file or the gap demo files of vue-flicking as well?

WoodNeck:
@malangfox
It's okay to remove the obsolete demos.

PR #747, opened 2022-11-15
docs: update readme.md about local installation

wlsdn2749:
## Details

Update readme.md to prevent this build problem
https://github.com/naver/egjs-flicking/discussions/746

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -2,6 +2,16 @@
 
 This website is built using [Docusaurus 2](https://v2.docusaurus.io/), a modern static website generator.
 
+
+## Prerequisites
+```console
+npm install
+npm run docs:build
+```
+
+Before you start a local development, Make sure the required command in the root package to generate an api file. If not, You may encounter "Not Found Error" regarding an api file.
+
+
 ## Installation
 
 ```console


malangfox (approved):
LGTM

docs/README.md
```diff
@@ -2,6 +2,16 @@
...
 This website is built using [Docusaurus 2](https://v2.docusaurus.io/), a modern static website generator.
 
+
+## Prerequisites
+```console
+npm install
+npm run docs:build
+```
+
+Before you start a local development, Make sure the required command in the root package to generate an api file. If not, You may encounter "Not Found Error" regarding an api file.
```
WoodNeck, line 12: How about using more simple explanation like this:
```
Before starting a local development, run `docs:build` in the root directory to generate API documents.
```

WoodNeck (commented):
Thanks for the suggestion, but this file(`docs/README.md`) is generated from Docusaurus by default and we didn't change it at all.
In other words, we're not efficiently using this file at the moment.

So my suggestion is, to prevent documents/guides from being spread, how about updating the [CONTRIBUTING.md](https://github.com/naver/egjs-flicking/blob/master/CONTRIBUTING.md) file instead?

docs/README.md
wlsdn2749: Oh, I didn't know that, That is more simple than mine, I'll do re - Pull requests with same simple explanations in  the [CONTRIBUTING.md](https://github.com/naver/egjs-flicking/blob/master/CONTRIBUTING.md) file, thank you for your reply.

PR #748, opened 2022-11-17
docs: I updated width of post elements in css file of showcases

chayou20:
post panel main be showed with next panel element.
because .post-panel.main width: is 88% in css file

I changed this property value to 100%

.posts .flicking .post-panel.main {
  position: relative;
  z-index: 2;
  border-left: 0;
  border-right: 0;
  width: 100%;
}


WoodNeck:
Thanks for the PR, but that's intended.

PR #753, opened 2022-11-17
Camera mode docs translate into korean

yongilkim01:
## Issue

Translate camera mode docs into korean.


WoodNeck:
As I said in #751

```
You should change jsdoc comments in our src files, not markdown files.
```

Our API documents are all auto-generated, changing the markdown file takes no effect.

---

Also, [CameraMode](https://github.com/naver/egjs-flicking/blob/master/src/camera/mode/CameraMode.ts) is not documented at the moment, that's just a brief placeholder.
Your approach should be documenting the actual class of its purpose, plus explanations of methods, etc.
We definitely need help improving our documents, but this PR actually takes no effect in terms of improvement.

Please reopen after fixing.

PR #762, opened 2022-12-05
docs(NavigationToOptions): Update API document

wlsdn2749:
Update Link to the option was added to the property section of the api document.

## Details

There are many parts similar to options in the property section of the api document,
but I thought it would be nice to have a connection with options with a simple ui.


malangfox:
Thank you for your suggestions for improving our documentation.
However, our api documentation files are automatically generated.
You need to change the jsdoc comments inside the src files to change our api documentation.
You can check the changed documentation with local environment by `npm run docs:build` after changing the jsdoc comments.

wlsdn2749:
Oh, I just know that api documentation files are automatically generated by docusaurus (right?)
then.. Before changing comments inside src file, I would like to link an api document to Option Tab like `see` without a http `link`

After a little modification, only the imported items from Flicking could be moved directly without a link. Is there any other way to do that?

plus, only api document file is automatically generated? Others like options or showcases don't seem to be automatically generated.

Issue #765, opened 2022-12-07
Overlapping components of Showcases Page


chayou20:
## Description

![1](https://user-images.githubusercontent.com/48912304/206247668-b5e104eb-927f-4d50-b30c-2e38d1169bf5.PNG)
![4](https://user-images.githubusercontent.com/48912304/206247732-ac1b99c5-bfed-476f-8e60-883725c5f4ad.PNG)

I find overlapping components of Showcases Page.

![image](https://user-images.githubusercontent.com/48912304/206248084-7ef0e2a0-d097-43a0-abab-f1c843790cfe.png)

![image](https://user-images.githubusercontent.com/48912304/206248592-d26e78a3-a327-4f6b-997f-2cf6ab694fe0.png)

and checked z-index of .navbar--fixed-top and artwindow-overlay

'--ifm-z-index-fixed' value is 9 and 'z-index of artwindow-overlay' is 12 on this page

I checked this source code in local but It has different value

'--ifm-z-index-fixed' is 200 and 'z-index of artwindow-overlay' is 2001.

![image](https://user-images.githubusercontent.com/48912304/206250587-580cf074-3487-4063-a27a-be8d705d0973.png)

I tried to modify the file in the path above, but it didn't work.

I think that changed on a page that actually works.

So I added the code below to the custom.css file and modified it.

![image](https://user-images.githubusercontent.com/48912304/206251337-122ad4d4-a4fc-411f-8445-4d7601752dd8.png)

before

![bandicam-2022-12-08-01-52-40-286](https://user-images.githubusercontent.com/48912304/206252906-86ef5d15-b918-4150-8968-ed0c205cb7c7.gif)

after

![bandicam-2022-12-08-01-52-56-284](https://user-images.githubusercontent.com/48912304/206252648-bce92ba4-2ced-46d9-835e-4c864c553c90.gif)

malangfox:
This issue has been resolved by releasing version 4.10.6 of the document.
Thank you for improving our page.

PR #766, opened 2022-12-07
Overlapping of Components

chayou20:
## Description

I find overlapping components of Showcases Page.

image

image

and checked z-index of .navbar--fixed-top and artwindow-overlay

'--ifm-z-index-fixed' value is 9 and 'z-index of artwindow-overlay' is 12 on this page

I checked this source code in local but It has different value

'--ifm-z-index-fixed' is 200 and 'z-index of artwindow-overlay' is 2001.

image

I tried to modify the file in the path above, but it didn't work.

I think that changed on a page that actually works.

So I added the code below to the custom.css file and modified it.

image

before

bandicam-2022-12-08-01-52-40-286

after

bandicam-2022-12-08-01-52-56-284

### DIFF
--- a/docs/src/css/custom.css
+++ b/docs/src/css/custom.css
@@ -38,6 +38,9 @@ html[data-theme='dark'] {
 .navbar__brand {
   padding: var(--ifm-navbar-item-padding-vertical) 1rem;
 }
+.navbar--fixed-top {
+  z-index: 2002;
+}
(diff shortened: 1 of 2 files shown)


malangfox (approved):
LGTM

malangfox:
This PR has been closed with no updates.
Issue #765 has been resolved by releasing version 4.10.6 of the document.

PR #771, opened 2022-12-29
fix: prevent invalid coordinate value in AxesController

malangfox:
## Details
When using a specific combination of options such as below, the internal position exceeds the range of coordinates.
```
{
  circular: true,
  align: "prev",
  duration: 0,
  moveType: "strict"
}
```
Here is the [reproduction demo](https://codepen.io/malangfox/pen/QWByPgp) of this issue.
When you click the panel, the internal coordinates and range change as shown below at the last point,

image

And when you interact with Flicking once more, it shows an abnormal movement to return this value to 0 in the range.

This was because the `animateTo` in AxesController updates the range and position one more time separately from the strict control update range and position when duration is 0.

According to what I have checked, only the strict type has the behavior of changing the pos and range of the Axes instance when the panel is changed.
```ts
  public setActive = (newActivePanel: Panel, prevActivePanel: Panel | null, isTrusted: boolean) => {
    super.setActive(newActivePanel, prevActivePanel, isTrusted);
    this.updateInput();
  };
```
So, I removed unnecessary behavior of `animateTo` in strict move type.

### DIFF
--- a/src/control/AxesController.ts
+++ b/src/control/AxesController.ts
@@ -8,6 +8,7 @@ import Flicking from "../Flicking";
 import FlickingError from "../core/FlickingError";
 import * as AXES from "../const/axes";
 import * as ERROR from "../const/error";
+import { MOVE_TYPE } from "../const/external";
 import { circulatePosition, getFlickingAttached, parseBounce } from "../utils";
 import { ControlParams } from "../type/external";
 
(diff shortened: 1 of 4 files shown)


src/control/AxesController.ts
```diff
@@ -378,15 +379,16 @@ class AxesController {
...
-        : position;
-
-      axes.axisManager.set({ [AXES.POSITION_KEY]: newPos });
+      if (flicking.moveType !== MOVE_TYPE.STRICT) {
+        const camera = flicking.camera;
+        const newPos = flicking.circularEnabled
+          ? circulatePosition(position, camera.range.min, camera.range.max)
+          : position;
+        axes.axisManager.set({ [AXES.POSITION_KEY]: newPos });
+      }
```
WoodNeck, line 391: It's a good practice to not include some child-class-specific code inside the general loop.

So, my suggestion is:
As this code is quite specific for `moveType: "strict"`, what about updating `position` after this in `StrictControl` instead?
One good place to put it is `_animateToPosition` I guess, you can try to inherit it & check whether the `position` is inside the correct range and update it.

malangfox:
The cause of this issue was AxesController's `animateTo` that calculates invalid newPos when `duration` is 0.
However, I have checked that the implementation of the zero duration case of `animateTo` is unnecessary.
`Axes.setTo` used in `animateTo` changes the coordinates immediately without animation processing when the `duration` is 0, and the [finish event for animationFinishHandler is normally triggered](https://github.com/naver/egjs-axes/blob/master/packages/axes/src/animation/AnimationManager.ts#L270).

It also fixes one unit test that was misbehaving. It tries to change the position before Flicking's ready event occurs.

WoodNeck (approved):
LGTM

daybrush (approved):
LGTM

PR #777, opened 2023-01-25
fix: control fallback when interrupted

plrs9816:
## Issue
#776

## Details
When control is set to SnapControl or StrictControl, click event during panel sliding triggers restore behavior.
``` js
// src/control/SnapControl.ts > SnapControl > moveToPosition

// Restore to active panel
return this.moveToPanel(activeAnchor.panel, {
  duration,
  axesEvent
});
```
To reproduce, go to [official demo](https://naver.github.io/egjs-flicking/ko/Demos), drag fast, then click during sliding. When mouseup, the panel restores to starting position.

However this behavior in most circumstances are rather confusing to end user since the camera has already moved far from original panel but it restores to original position.
So it would feel more natural for camera to slide to nearest panel when interrupted.
``` js
// src/control/SnapControl.ts > SnapControl > moveToPosition

// Fallback to nearest panel from current camera
return this.moveToPanel(anchorAtCamera.panel, {
  duration,
  axesEvent
});
```
## Caveats
It passed the unit test but I'm not sure if changing restore behavior will cause other side effects rather than click interrupt during sliding.

### DIFF
--- a/src/control/StrictControl.ts
+++ b/src/control/StrictControl.ts
@@ -242,8 +242,8 @@ class StrictControl extends Control {
       targetPanel = adjacentAnchor!.panel;
       targetPos = adjacentAnchor!.position;
     } else {
-      // Restore to active panel
-      targetPos = camera.clampToReachablePosition(activePanel.position);
+      // Fallback to nearest panel from current camera
+      targetPos = camera.clampToReachablePosition(anchorAtPosition.position);
       targetPanel = activePanel;
     }
 
--- a/src/control/StrictControl.ts
+++ b/src/control/StrictControl.ts
@@ -243,8 +243,14 @@ class StrictControl extends Control {
       targetPos = adjacentAnchor!.position;
     } else {
       // Fallback to nearest panel from current camera
-      targetPos = camera.clampToReachablePosition(anchorAtPosition.position);
-      targetPanel = activePanel;
+      const anchorAtCamera = camera.findNearestAnchor(camera.position);
+      if (!anchorAtCamera) {
+        return Promise.reject(new FlickingError(ERROR.MESSAGE.POSITION_NOT_REACHABLE(position), ERROR.CODE.POSITION_NOT_REACHABLE));
+      }
+      return this.moveToPanel(anchorAtCamera.panel, {
+        duration,
+        axesEvent
+      });
     }
 
     this._triggerIndexChangeEvent(targetPanel, position, axesEvent);
--- a/test/unit/control/SnapControl.spec.ts
+++ b/test/unit/control/SnapControl.spec.ts
@@ -291,6 +291,7 @@ describe("SnapControl", () => {
 
         const control = flicking.control;
         const camera = flicking.camera;
+        const moveSpy = sinon.spy(control, "moveToPanel");
 
         // Suppress animation interrupt error
         const promise = flicking.moveTo(2, 1500).catch(error => error);
@@ -301,6 +302,7 @@ describe("SnapControl", () => {
         tick(1000);
         await promise;
 
+        expect(moveSpy.calledTwice).to.be.true;
         expect(control.activePanel.index).to.equal(camera.findNearestAnchor(position).panel.index);
       });
     });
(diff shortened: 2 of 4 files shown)


malangfox (approved):
LGTM.

If it is possible to add a unit test for this change, it would be nice to add it to the tests of SnapControl and StrictControl.
Otherwise, we will add them after this change is applied.

plrs9816:
Added unit test, and StrictControl was not working as expected so also fixed that.

src/control/StrictControl.ts
```diff
@@ -242,9 +242,15 @@ class StrictControl extends Control {
...
       targetPos = adjacentAnchor!.position;
     } else {
-      // Restore to active panel
-      targetPos = camera.clampToReachablePosition(activePanel.position);
-      targetPanel = activePanel;
+      // Fallback to nearest panel from current camera
+      const anchorAtCamera = camera.findNearestAnchor(camera.position);
+      if (!anchorAtCamera) {
+        return Promise.reject(new FlickingError(ERROR.MESSAGE.POSITION_NOT_REACHABLE(position), ERROR.CODE.POSITION_NOT_REACHABLE));
+      }
```
WoodNeck, line 249: nit: `findNearestAnchor` always returns a value if at least one panel exists, so basically there're no cases that `anchorAtCamera` doesn't exist.

test/unit/control/SnapControl.spec.ts
```diff
@@ -282,6 +282,27 @@ describe("SnapControl", () => {
...
+        // Suppress animation interrupt error
+        const promise = flicking.moveTo(2, 1500).catch(error => error);
+        tick(500);
+        const position = camera.position;
+        // Simulate interrupt
+        await simulate(flicking.element, { deltaX: 0, duration: 100 }, 1000);
+        tick(1000);
+        await promise;
+
+        expect(control.activePanel.index).to.equal(camera.findNearestAnchor(position).panel.index);
```
WoodNeck, line 304: Can you add a `expect` that `control.activePanel.index` not equals `2`? To prove that it's correctly interrupted.

WoodNeck (approved):
LGTM

src/control/StrictControl.ts
plrs9816: camera.findNearestAnchor method returns nullable, but moveToPanel does not accept nullable params.
Is it okay to non-null assert the return?

test/unit/control/SnapControl.spec.ts
plrs9816: (not in this test case, but...) Index can be 2 even when interruption occurs if camera position is very close to last panel... so I didn't add that.

src/control/StrictControl.ts
WoodNeck: You're right, that's typed the same as it behaves.
I think it's okay just as it is, in both cases, they'll throw errors but it's easier to find such circumstances when we throw errors like that.

test/unit/control/SnapControl.spec.ts
plrs9816: (not in this test case, but...) Index can be 2 even when interruption occurs if camera position is very close to last panel...
So checking if the fallback behavior was triggered seems more appropriate.

WoodNeck: That's weird, maybe the element's width is wrong, or something is wrong with `tick` in that case?
Please report to me for that case when it occurs again, I'll try to fix that.

Another approach could be listening to `move` event and then interrupting when the `nearestAnchor` is panel 1.
Anyway, I think it's fine for now 👌

plrs9816: Isn't it normal behavior to do so? I mean, when the sliding almost ended and camera is closest to the destination slide (slide 2), the fallback should `moveTo` 2.

WoodNeck: The problem is, we're using sinon's fake timer, and constant time for `tick` and `simulate`.
The result should be consistent in that kind of environment. If not, there must be something that we're missing.
For example, there was a case where some tests were not removing elements after the test so that affected other test cases depending on the test execution order. We have to exclude such cases.

plrs9816: Ah I see what you mean... there's no problem, the results are consistent. I meant just wrote that way for genereal cases.

daybrush (approved):
LGTM

PR #782, opened 2023-02-02
fix: apply transform after elements are fully rendered

### LINKED ISSUE #773: cirular option makes screen flickers
## Description

i use Flicking library in **next JS** project , when i use **circular option** and it makes screen flickers, when first rendered.
when commenting out **circular option**, it is not flicking.
 but circular option is absolutely necessary. is there way to not flicking?

[https://www.galmaetgil.kr/](url)

스크린샷 2023-01-06 오후 3 31 57
>> these image slide is used circular option
스크린샷 2023-01-06 오후 3 31 18
>> these image is my code

WoodNeck:
## Issue
#773 

## Details
This fixes #773 by delaying CSS transform application until the elements are fully rendered in the correct order.
This can fix panels flickering on the first render in the SSR frameworks.

### DIFF
--- a/packages/ngx-flicking/projects/ngx-flicking/src/lib/NgxRenderer.ts
+++ b/packages/ngx-flicking/projects/ngx-flicking/src/lib/NgxRenderer.ts
@@ -36,7 +36,7 @@ class NgxRenderer extends ExternalRenderer {
     strategy.renderPanels(flicking);
 
     this._resetPanelElementOrder();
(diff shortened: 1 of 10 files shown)


daybrush:
It seems that everything, including vanilla, can be applyTransform, so I wonder if we can make it common.

daybrush (approved):
LGTM

malangfox (approved):
LGTM

Issue #770, opened 2022-12-28
Vue 3 + vite warnings


preetamslot:
I use Quasar and in Webpack all worked fine, but with Vite the carousel won't start.
it gives me these warnings in the console:
```
Flicking.ts:160 
[Vue warn]: Property "vanillaFlicking" was accessed during render but is not defined on instance. 
  at <Flicking plugins= [Arrow2] ref="carousel" options= {align: 'prev', circular: false, bound: true, useResizeObserver: true, resizeDebounce: 100, …}  ... >  
```

```
Flicking.ts:186 
[Vue warn]: Property "getPanels" was accessed during render but is not defined on instance. 
  at <Flicking plugins= [Arrow2] ref="carousel" options= {align: 'prev', circular: false, bound: true, useResizeObserver: true, resizeDebounce: 100, …}align: "prev"bound: truecircular: falserenderOnlyVisible: trueresizeDebounce: 100useResizeObserver: true[[Prototype]]: Object  ... > 

```

If I console log the carousel ref:
```
{
    "viewportTag": "div",
    "cameraTag": "div",
    "cameraClass": "",
    "hideBeforeInit": false,
    "options": {
        "align": "prev",
        "circular": false,
        "bound": true,
        "useResizeObserver": true,
        "resizeDebounce": 100,
        "renderOnlyVisible": true
    },
    "plugins": [
        {
            "_flicking": null,
            "_parentEl": null,
            "_prevElSelector": ".flicking-arrow-prev",
            "_nextElSelector": ".flicking-arrow-next",
            "_disabledClass": "flicking-arrow-disabled",
            "_moveCount": 1,
            "_moveByViewportSize": false
        }
    ]
}
```

Any pointer on how to get this working?
Thanks

malangfox:
Hello @preetamslot.

We would like to see a detailed implementation of how you're using our component.
In Vue, the declaration must be made through methods and data.
If you declare it in created, the warning attached to this issue can occur.

malangfox:
Hello @preetamslot.

This seems to be an issue that needs to be fixed in our components.
Can we check if there is a repo that reproduces this issue with the environment?

After we release the patch that fixed this issue, we will give you additional comments.

malangfox:
@preetamslot 

Can we check the version of the packages you are using as we further investigate the circumstances of this issue?
When we tried to reproduce this error, the issue you left was not reproduced when using 3.2.45 version of Vue and 4.0.0 version of Vite.

[...]

PR #790, opened 2023-03-17
refactor: declare methods in methods instead of created

malangfox:
## Issue
#770

## Details
This changes `getPanels` and `getVirtualPanels` to be initialized in `methods` instead of `created`.
In some cases, such a warning appears because the methods declared in `created` are not recognized.

### DIFF
--- a/packages/vue3-flicking/src/Flicking.ts
+++ b/packages/vue3-flicking/src/Flicking.ts
@@ -42,54 +42,6 @@ const Flicking = defineComponent({
     this.renderEmitter = new Component();
     this.diffResult = null;
 
-    this.getPanels = () => {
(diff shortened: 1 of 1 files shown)


WoodNeck (commented):
Have you been able to test this fix works in both our demo & Vite?
I remember that was defined like that because `getCurrentInstance` returned the wrong result when used in methods I guess.

I don't know if it was an issue with a specific version of Vue, but please at least test whether it works fine with the latest version of Vue.

daybrush (approved):

daybrush:
LGTM

@malangfox 
Check if this is the case in vue2 as well.

daybrush (approved):
LGTM

PR #796, opened 2023-04-03
fix: fix setStatus not working when index is 0

malangfox:
## Issue
#795 
A [demo](https://codepen.io/malangfox/pen/QWVebpj) that reproduces the issue.

## Details
This is a bug caused by `if (index)` is false when index is 0.

### DIFF
--- a/src/Flicking.ts
+++ b/src/Flicking.ts
@@ -1397,7 +1397,7 @@ class Flicking extends Component<FlickingEvents> {
       renderer.batchInsert({ index: 0, elements: parseElement(panels.map(panel => panel.html!)), hasDOMInElements: true });
     }
 
-    if (index !== undefined) {
+    if (index != null) {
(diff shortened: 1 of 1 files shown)


src/Flicking.ts
```diff
@@ -1397,7 +1397,7 @@ class Flicking extends Component<FlickingEvents> {
       renderer.batchInsert({ index: 0, elements: parseElement(panels.map(panel => panel.html!)), hasDOMInElements: true });
     }
 
-    if (index) {
+    if (index !== undefined) {
```
WoodNeck, line 1400: nit) `index != null` would be better for general I think.
As it can both exclude `null` and `undefined`.

WoodNeck (approved):
LGTM

daybrush (approved):
LGTM

Issue #815, opened 2023-07-06
ngx-flicking 4.10.10 is broken directive styles under Angular V16


wen9218:
## Description
Hi, ngx-flicking 4.10.10 is not working properly under ng v16. directive styles like flicking-camera, flicking-panels are broken.
But ngx-flicking 4.10.10-beta.0 works well.
Overall
- ngx-flicking 4.10.11-beta0 - styles are **broken**
- ngx-flicking 4.10.10 - styles are **broken**
- ngx-flicking 4.10.10-beta0 - styles work well.

Can you please check and fix it in the next version?

born2net:
Hi... yes we are having the same issues, please let us know if you need more input,
thank you.

daybrush:
@wen9218 @born2net 

I'll check. It seems that the problem occurred when the angular version of the component was upgraded.

daybrush:
@born2net @wen9218 

Can you tell me how Flicking was imported and how it was used in the template?

wen9218:
@daybrush, I imported ngx-flicking module and used like this. Thanks for advancing.

![image](https://github.com/naver/egjs-flicking/assets/57497730/83d5e56a-57f2-4e16-a1ed-f2fc36550f19)
![image](https://github.com/naver/egjs-flicking/assets/57497730/2c3b495e-0f0c-4301-a77b-7596875d4c97)

malangfox:
Hello @wen9218.

We have released @egjs/ngx-flicking version 4.10.11.
Can you check if the issue is reproduced in our latest version?

wen9218:
Hi, @daybrush, @malangfox, 
I checked v4.10.11. It works fine. Thanks a lot!

born2net:
thank you!!!

PR #816, opened 2023-07-06
fix: add missing styleUrls in ngx-flicking

malangfox:
## Issue
#815

### DIFF
--- a/packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
+++ b/packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
@@ -72,6 +72,9 @@ import NgxElementProvider from './NgxElementProvider';
     class: 'flicking-viewport',
     style: 'display: block;',
   },
+  styleUrls: [
+    "../../node_modules/@egjs/flicking/dist/flicking.css"
+  ],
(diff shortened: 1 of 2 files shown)


packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
```diff
@@ -72,6 +72,9 @@ import NgxElementProvider from './NgxElementProvider';
     class: 'flicking-viewport',
     style: 'display: block;',
   },
+  styleUrls: [
+    "../../node_modules/@egjs/flicking/dist/flicking.css"
```
daybrush, line 76: Could it be this way?

`@egjs/flicking/dist/flicking.css`

daybrush (approved):
LGTM

packages/ngx-flicking/projects/ngx-flicking/src/lib/ngx-flicking.component.ts
malangfox: I've checked that absolute path `@egjs/flicking/dist/flicking.css` doesn't work with the intended path when building the package.
https://github.com/angular/angular-cli/issues/20162

Issue #835, opened 2023-11-06
Direction not working correctly for RTL


huynl-96:
## Description

I'm using `vue-flicking` for my work. The `horizontal` property is only working for normal LTR flow.
For RTL, the animation is not correct and not showing the next panels

## Steps to check or reproduce

https://github.com/naver/egjs-flicking/assets/15758406/42232bb0-8713-426f-a99c-990e4be64ba2

malangfox:
This issue is resolved in version 4.11.1 of @egjs/flicking.
Try the latest version of the package and leave a comment if you encounter additional issues.

PR #836, opened 2023-11-08
fix: fix broken layout when direction rtl is applied

malangfox:
## Issue
#835 

## Details
This issue fixes the layout breaking when `direction: rtl` is applied to the viewport.


daybrush (approved):
LGTM

src/Flicking.ts
```diff
@@ -1511,6 +1520,7 @@ class Flicking extends Component<FlickingEvents> {
     camera.updateAdaptiveHeight();
     camera.updateOffset();
     await renderer.render();
+    this._updatePanelOrder(camera);
```
daybrush, line 1523: I think it would be better to move to the camera class method.

malangfox: I have modified `updatePanelOrder` to camera class method.
You can check the most recent commit.
`updatePanelOrder` requires `flicking.control.controller.updateDirection();` to be called.
Should this be the camera class method, even though it interferes with flicking's controller?

Issue #844, opened 2024-01-10
threshold option is not working for specific move direction


malangfox:
## Description
If you use the `threshold` option with `horizontal: false` and `align: "prev"` options, the `threshold` option will not work for specific move direction.

## Steps to check or reproduce
[https://codepen.io/malangfox/pen/NWJNbje](https://codepen.io/malangfox/pen/NWJNbje)
This issue can be reproduced if you move to panel 2 or 3 and drag upside less than 100px (threshold).

PR #845, opened 2024-01-10
fix: fix threshold option not applying snap threshold

malangfox:
## Issue
#844

## Details
- When determining the `snapThreshold` in `_calcSnapThreshold`, the value of the `threshold` option for the Flicking instance must be applied.
- We didn't have a unit test that used vertical Flicking (DEFAULT_VERTICAL in a unit test) and this is the first time we've used it, I fixed it to be similar to the Flicking for other tests.


daybrush (approved):
LGTM

Issue #850, opened 2024-02-08
Position Error occured when using bound and custom align props together.


yungo1846:
## Description
Hello I've met error when using bound and custom align props together.
<img width="502" alt="image" src="https://github.com/naver/egjs-flicking/assets/59409762/9c768818-86e7-44ac-b320-2ecf774a1b6e">

```tsx
import { forwardRef } from "react";

import { Arrow } from "@egjs/flicking-plugins";
import Flicking, { ViewportSlot } from "@egjs/react-flicking";

import "@egjs/flicking-plugins/dist/arrow.css";
import "@egjs/react-flicking/dist/flicking.css";

interface Props {
  children: React.ReactNode;
}

const plugins = [new Arrow({})];

export const Carousel = forwardRef<Flicking, Props>(
  ({ children }: Props, ref) => {
    return (
      <Flicking
        ref={ref}
        bound
        align={{ panel: "120px", camera: "50%" }}
        defaultIndex={0}
        moveType={["snap"]}
        plugins={plugins}
      >
        {children}
        <ViewportSlot>
          <span className="flicking-arrow-prev"></span>
          <span className="flicking-arrow-next"></span>
        </ViewportSlot>
      </Flicking>
    );
  }
);

```

## Steps to check or reproduce
[...]

PR #851, opened 2024-02-15
fix: add findActiveAnchor fallback for bound flicking

malangfox:
## Issue
#850

https://codepen.io/malangfox/pen/WNmPemg
Pressing the moveTo button, opening the console, and dragging the flicking should reproduce it.

## Details

Position is not reachable error occurs when using the `moveTo` method for the first panel with `bound: true`.

image

In the above demo, when 4 Panels are placed, the total number of Anchors is 2.

This is because the Panel labeled 1 and the Panel labeled 2 share the same Anchor point.
Similarly, the panel with 3 and the panel with 4 share the same anchor point.

This is because there is no way to send Panels 1 / 4 to the center with the bound option enabled, which removes the left and right empty spaces.

This means that in this situation, only the Anchor points of Panel 2 and Panel 3 exist.

If we use the `moveTo` method with Panel 1, the Anchor point we stop at is the Anchor of Panel 2, but the index of the active Panel in the flicking will be the first Panel.

In this situation, `moveToPosition`, which calculates the arrival point at the end of a drag with `moveType: "snap"`, has the following logic.

```ts
const activeAnchor = camera.findActiveAnchor();
const anchorAtCamera = camera.findNearestAnchor(camera.position);
if (!activeAnchor || !anchorAtCamera) {
  return Promise.reject(new FlickingError(...));
}
```

and `findActiveAnchor` behaves as follows.

```ts
public findActiveAnchor(): AnchorPoint | null {
  const flicking = getFlickingAttached(this._flicking);
  const activeIndex = flicking.control.activeIndex;

  return find(this._anchors, anchor => anchor.panel.index === activeIndex);
}
```

The `activeIndex` is Panel 1, but none of the anchors corresponds to Panel 1, so it returns null.

Since `activeAnchor` is null, an error message is raised via Promise.reject.

I handled the exception of the `findActiveAnchor` by returning the closest Anchor to the active panel's position as a fallback if no corresponding Anchor for the active panel exists.

### DIFF
--- a/src/camera/Camera.ts
+++ b/src/camera/Camera.ts
@@ -404,9 +404,11 @@ class Camera {
    */
   public findActiveAnchor(): AnchorPoint | null {
     const flicking = getFlickingAttached(this._flicking);
-    const activeIndex = flicking.control.activeIndex;
+    const activePanel = flicking.control.activePanel;
 
-    return find(this._anchors, anchor => anchor.panel.index === activeIndex);
+    if (!activePanel) return null;
+
+    return find(this._anchors, anchor => anchor.panel.index === activePanel.index) ?? this.findNearestAnchor(activePanel.position);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


daybrush (approved):
LGTM

Issue #854, opened 2024-03-12
Events returns wrong direction on circular when calling prev() on first panel or next() on last panel.


fully-john:
## Description
I have a circular flicking with strict moveType that consists of three panels, all of them full screen.

I have buttons to call next() and previous() as well as the normal swipe motion.
However, when I'm on the first panel (index 0) and trigger prev(), the direction in WillChangeEvent.direction is "NEXT". However, if it's triggered by a swipe, the event returns the correct direction (in this case: "PREV"). The same happens in the opposite side of the panels. e.g. when I'm on the last panel and call next(), the WillChangeEvent.direction is "PREV". If triggered by swipe, the direction will be correct.

## Steps to check or reproduce
The following code returns wrong direction on .prev() at first panel and next() on last panel:
```
<SomeButton onClick={() => await flickRef.current.prev()} />
<Flicking
          ref={flickRef}
          circular
          duration={600}
          threshold={40}
          align={"center"}
          onWillChange={(e: WillChangeEvent) => {
            console.log("WILL CHANGE", e);
            flickStart(e.index, e.direction);
          }}
          onChanged={(e: ChangedEvent) => {
            console.log("DID CHANGE", e);
            flickEnd(e.index);
          }}
          moveType={"strict"}
          changeOnHold={true}
        >
...
</Flicking>
```

Also, thanks a bunch for all the work you put in. Great package overall!

malangfox:
Hello @fully-john.

We have reproduced the issue you left and this seems to be a bug in our component.
We will leave additional comments when we release a patch that fixes this issue.

malangfox:
We have released @egjs/flicking 4.11.3-beta.2, which fixes this issue.
Can you try the above beta version and see if it fixes this issue?

PR #856, opened 2024-03-18
fix: fix direction of willChange event when circular

malangfox:
## Issue
#854 

## Details
This fixes the behavior of `getDirection`, fixing cases where the direction of the `willChange` event is incorrect.


daybrush (approved):
LGTM

PR #868, opened 2024-06-21
fix: Add check before update

ishmalez:
## Details
Sometimes when mounting and unmounting different instances of Flicking slider, variable this.slotDiffer is undefined, throwing an exception on beforeUpdate() execution.
This fix is just to avoid the exception.


malangfox (approved):
LGTM

RodrigoProjects:
Thanks for the fix! Will this be in the next minor @malangfox ?

PR #884, opened 2024-08-21
feat: add resizeObservePanel option

### LINKED ISSUE #883: autoResize option cannot detect changes in Panel size
## Description
Flicking with `autoResize` option does not resize when the Panel changes size and the viewport remains the same size.

[Demo](https://codepen.io/malangfox/pen/BagYwvx)

malangfox:
## Issue
#883 

## Details
This fixes #883 by adding `resizeObservePanel` option to Flicking instance.

### DIFF
--- a/src/core/AutoResizer.ts
+++ b/src/core/AutoResizer.ts
@@ -63,6 +63,13 @@ class AutoResizer {
     return this;
   }
 
+  public unobservePanels(): this {
+    this._flicking.panels.forEach(panel => {
+      this.unobserve(panel.element);
+    });
+    return this;
+  }
+
   public observe(element: HTMLElement): this {
     const resizeObserver = this._resizeObserver;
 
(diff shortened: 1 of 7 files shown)


src/renderer/Renderer.ts
```diff
@@ -335,6 +335,15 @@ abstract class Renderer {
     // Update camera & control
     this._updateCameraAndControl();
 
+    if (flicking.autoResize && flicking.useResizeObserver) {
```
daybrush, line 338: Add defense code to check if element exists.

src/core/AutoResizer.ts
```diff
@@ -36,7 +40,11 @@ class AutoResizer {
         ? new ResizeObserver(this._skipFirstResize)
         : new ResizeObserver(this._onResize);
 
-      resizeObserver.observe(flicking.viewport.element);
+      [
```
daybrush, line 43: Is it possible to cancel the schedule when calling resize with a method?

Issue #905, opened 2024-12-13
State is not updated when panel's className changes


malangfox:
## Description
State is not updated when children's className changes

## Steps to check or reproduce
[Demo](https://codesandbox.io/p/sandbox/react-flicking-classname-state-update-9thqxt)

Pressing the new Color button does not immediately apply the changed class.
You have to change the browser size to fire `reisze()` to apply changed class.

PR #906, opened 2024-12-13
fix: check other props for react state change

malangfox:
## Issue
https://github.com/naver/egjs-flicking/issues/905

## Details
Checks all props when determining whether the panel's state is updated.


daybrush (approved):

packages/react-flicking/src/react-flicking/Flicking.tsx
```diff
@@ -267,7 +267,7 @@ class Flicking extends React.Component<Partial<FlickingProps & FlickingOptions>>
       const nextChild = nextChildren[idx];
 
       if (child.key && nextChild.key) {
-        return child.key === nextChild.key;
+        return child.key === nextChild.key && Object.keys(child.props).every(key => child.props[key] === nextChild.props[key]);
```
daybrush, line 270: Actually, this has no meaning. The cases or objects of children are easy to change.

PR #909, opened 2025-02-04
Fix deprecated findDOMNode usage in React 19

jtwite:
Update the library to remove the use of deprecated `findDOMNode` and replace it with a ref-based approach.

* **NonStrictPanel.tsx**
  - Replace `findDOMNode` with `React.createRef` to access the DOM element.
  - Update the `nativeElement` getter to return the current ref.
  - Modify the `render` method to clone the element with the ref.

* **ReactElementProvider.ts**
  - Update the constructor to accept a ref instead of a `StrictPanel` or `NonStrictPanel`.
  - Update the `element` getter to return the current ref.
  - Modify the `show` and `hide` methods to use the ref.

* **Flicking.tsx**
  - Remove the `useFindDOMNode` prop and use a ref-based approach.
  - Update the `_createPanelRefs` method to use `React.createRef`.
  - Update the `_getPanels` method to use the new ref-based approach.

* **ReactRenderer.ts**
  - Update the `_collectPanels` method to use the new ref-based approach.
  - Update the `_createPanel` method to use the new ref-based approach.

* **package-lock.json**
  - Update the version to `4.12.0-snapshot`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "@egjs/flicking",
-	"version": "4.11.4-snapshot",
+	"version": "4.12.0-snapshot",
 	"lockfileVersion": 2,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "@egjs/flicking",
-			"version": "4.11.4-snapshot",
+			"version": "4.12.0-snapshot",
 			"license": "MIT",
 			"dependencies": {
 				"@egjs/axes": "^3.9.1",
--- a/packages/react-flicking/src/react-flicking/Flicking.tsx
+++ b/packages/react-flicking/src/react-flicking/Flicking.tsx
@@ -332,9 +332,7 @@ class Flicking extends React.Component<Partial<FlickingProps & FlickingOptions>>
         : getRenderingPanels(vanillaFlicking, diff(origChildren, origChildren))
       : origChildren;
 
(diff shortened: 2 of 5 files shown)


jtwite:
Definitely!

malangfox:
`createRef` was added in later versions of React 16.
As long as we maintain support for early versions of React 16, we can't use `createRef` for this issue.

We are looking into other ways to resolve this issue, so feel free to give us your insight on this.

InJaEE:
> `createRef` was added in later versions of React 16. As long as we maintain support for early versions of React 16, we can't use `createRef` for this issue.
> 
> We are still looking into other ways to resolve this issue, so feel free to give us your insight here.

@malangfox 
Hello. I found that createRef is already being used in Flicking.tsx and StrictPanel.tsx. Could this be unrelated to React 16 compatibility?

malangfox:
@InJaEE 

Since we have confirmed that `createRef` already exists in our implementation, there should be no problem with this fix.

@jtwite 
Could you modify package.json and gitignore to match our main branch before merging?

cs09g:
@jtwite Please update PR to get merged 😊

malangfox:
Since there were no additional changes, We will apply the changes you left in PR https://github.com/naver/egjs-flicking/pull/931.

PR #918, opened 2025-04-23
Fix/resize logic

cjw783:
## Issue

Resize event occurs infinitely sometimes when panel height changes.

## Details

- Added `isResizing` flag to prevent resize event occurence while processing resize logic.
- Changed resize logic condition to make it occur only if there is difference between size before resize and size after resize.
- Added `optimizeSizeUpdate` option to skip unnecessary re-rendering of panels when resizing viewport.


src/core/AutoResizer.ts
```diff
@@ -33,14 +35,14 @@ class AutoResizer {
...
-        : new ResizeObserver(this._onResize);
+        ? new ResizeObserver((entries: ResizeObserverEntry[]) => this._skipFirstResize(entries))
+        : new ResizeObserver((entries: ResizeObserverEntry[]) => this._onResize(entries));
 
       resizeObserver.observe(flicking.viewport.element);
 
       this._resizeObserver = resizeObserver;
     } else {
-      window.addEventListener("resize", this._onResize);
+      window.addEventListener("resize", () => this._onResize([]));
```
daybrush, line 45: Please change the callback function of the event listener.

malangfox (approved):
LGTM

